PR #3, opened 2019-02-13
 add aio post and cheerlights examples

jerryneedell:
Added a few examples including use of settings.py
@ladyada  If you are working on better examples, feel free to close this PR -- I just thought I make them available.


jerryneedell:
Thanks! good catch -- I'll make those changes

ladyada:
i think this is good! we will want a ESPSPI_WiFiManager helper class at some point, which does the instantiation and link management for ya. it could go in this library or separate (probably best to keep in this one since its so dependant on the hardware init)

PR #4, opened 2019-02-15
Added WiFi Manager and changed examples

makermelissa:
I added a WiFi manager to handle a lot of the connection details and change the status light on the board. I also updated Jerry's examples to make use of the WiFi Manager.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -29,6 +29,8 @@ WiFi Manager for making ESP32 SPI as WiFi much easier
 * Author(s): Melissa LeBlanc-Williams, ladyada
 """
 
+# pylint: disable=no-name-in-module
+
 import neopixel
 import adafruit_esp32spi
 import adafruit_esp32spi.adafruit_esp32spi_requests as requests
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -26,11 +26,11 @@
 
 WiFi Manager for making ESP32 SPI as WiFi much easier
 
-* Author(s): Melissa LeBlanc-Williams
+* Author(s): Melissa LeBlanc-Williams, ladyada
 """
 
 import neopixel
-from adafruit_esp32spi import adafruit_esp32spi
+import adafruit_esp32spi
 import adafruit_esp32spi.adafruit_esp32spi_requests as requests
 
 class ESPSPI_WiFiManager:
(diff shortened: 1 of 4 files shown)


examples/esp32spi_cheerlights.py
```diff
@@ -4,87 +4,53 @@
...
-        for x in DATA_LOCATION:
-            value = value[x]
+        response = wifi.get(DATA_SOURCE)
+        print(response.json())
+        value=response.json()
+        for key in DATA_LOCATION:
+            value = value[key]
             print(value)
-        r.close()
+        response.close()
```
jerryneedell: With change to response, line 66 should also replace r with response

makermelissa: Oh, good catch

examples/esp32spi_cheerlights.py
```diff
@@ -4,87 +4,53 @@
...
 # neopixels
 pixels = neopixel.NeoPixel(board.A1, 16, brightness=0.3)
 pixels.fill(0)
-builtin = neopixel.NeoPixel(board.NEOPIXEL, 1, brightness=0.1)
-builtin[0] = 0
 
 # we'll save the value in question
 last_value = value = None
 the_time = None
 times = 0
```
jerryneedell: Not your change, but I don’t think times or the_time are used. Remove here and the increment of times below.

jerryneedell:
This looks great. I’d prefer to be able to test it before approval. I should be able to do that tomorrow morning.

makermelissa:
Ok done. I also added the examples to the RTD examples file.

makermelissa:
I decided to add params to the DocStrings just to polish it up a little more.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -0,0 +1,146 @@
...
+            try:
+                if self.debug:
+                    print("Connecting to AP...")
+                self.neo_status((100, 0, 0))
+                self._esp.connect_AP(bytes(self.ssid, 'utf-8'), bytes(self.password, 'utf-8'))
+                self.neo_status((0, 100, 0))
+            except (ValueError, RuntimeError) as error:
+                if self.debug:
+                    print("Failed to connect, retrying\n", error)
+                continue
```
jerryneedell: should this failure message be printed even if debug not enabled? As it is, it just fails silently and keeps retrying.

jerryneedell:
tested post and cheerlights examples on metro_m4_espress with Huzzah ESP32 and feather_nrt52840_express with Espressif DevKit  -- both worked great!

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
makermelissa: Yeah, I can remove the only show on debug condition

ladyada:
@kattni wanna bump this too?

kattni:
@ladyada @makermelissa has been keeping up with bumping libs like a champ! But I can take care of this one to get it done!

makermelissa:
Is bumping the same thing as pushing a release in this case?

ladyada:
ooh pardon - ok flip a coin :) yes when i say 'bump' i mean you increase the semver and release

kattni:
@makermelissa Yep! That's exactly what it is :)

kattni:
Already released!

makermelissa:
Ah, ok. Thanks.

PR #10, opened 2019-02-21
import either json or ujson

dhalbert:
We're renaming `ujson` to `json` in 4.0.0. Tries to import json parser as either `json` or `ujson` for upward compatibility. Renames import to `json_module` to avoid conflicts with variables named `json`.


makermelissa:
Ok tested good. It looks like it's working.

PR #12, opened 2019-02-23
change settings to secrets, add a localtime example, internally set 4K as max socket packet size

ladyada:


examples/esp32spi_aio_post.py
```diff
@@ -8,30 +8,32 @@
 
 print("ESP32 SPI webclient test")
 
-# Get wifi details and more from a settings.py file
+# Get wifi details and more from a secrets.py file
 try:
-    from esp32spi_settings import settings
+    from esp32spi_secrets import secrets
 except ImportError:
```
jerryneedell: Do you want the file to be esp32spi_settings.py or just settings.py?

examples/esp32spi_cheerlights.py
```diff
@@ -9,11 +9,11 @@
 import neopixel
 import adafruit_fancyled.adafruit_fancyled as fancy
 
-# Get wifi details and more from a settings.py file
+# Get wifi details and more from a secrets.py file
 try:
-    from esp32spi_settings import settings
+    from esp32spi_secrets import secrets
```
jerryneedell: from settings?

examples/esp32spi_secrets.py
```diff
@@ -1,7 +1,7 @@
 # This file is where you keep secret settings, passwords, and tokens!
```
jerryneedell: should this file be name secrets.py  -- or keep it here as esp32spi_secrets.py to distinguish from other examples but user should rename to secrets.py

Issue #13, opened 2019-02-26
adafruit_esp32spi_wifimanager: allow DotStar as status pixel?


anecdata:
Since ItsyBitsy M4 is a viable main MCU board to use with the ESP32 co-processor, would be nice to be able to use the built-in DotStar as the status pixel instead of adding an external NeoPixel.

ladyada:
yeah honestly we should just default to the built in LED - want to try a PR?

anecdata:
I'm 95% n00b at [Circuit]Python and 100% n00b at git[Hub] tools & process, wouldn't trust me to modify libraries at this point. Happy to test stuff though, I'm good at breaking code ;-)

anecdata:
Thought there'd be a simple abstraction or conditional possible, but DotStar not in n HAL or core modules. The CP RGB status LED works for both neos and dots though (CPX ibeing special), so maybe some abstraction there. Or... cheerlights uses fancyled?

jerryneedell:
@ladyada Are you suggesting using the builtin LED (D13 on most boards) instead of the neopixel? It would still be nice to pass in a Pin since there are some odd cases (particle boards for instance). But are you ok with just an on/off for the LED instead of the colors the neopixel now uses or do you want to preserve the option for a neopixel. Should we have 2 optional kwargs -- one for a Neopixel, one for an LED -- if None -- ignore them -- if set -- use them.

makermelissa:
I think this is a great idea! I didn't realize the ItsyBitsy had a DotStar. I'm going to think about the best way to implement this.

ladyada:
ideally you'd pass in a NeoPixel or DotStar *object* that you'd call `fill` on - no instantiation in the library itself

makermelissa:
Yeah, that's kind of along the lines of what I was thinking. That's about how I did the DotStar/NeoPixel abstraction in the FeatherWing library.

makermelissa:
Fixed #13 via #14.

PR #14, opened 2019-02-26
Abstracted Status NeoPixel to make DotStar usable

makermelissa:
I abstracted Status NeoPixel to make DotStar usable. I also updated all of the examples. As this change affects the Adafruit IO library, please do the PR shortly after this PR. I will link after it's created. This is in regards to issue #13.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -31,7 +31,6 @@ WiFi Manager for making ESP32 SPI as WiFi much easier
 
 # pylint: disable=no-name-in-module
 
-import neopixel
 import adafruit_esp32spi
 import adafruit_esp32spi.adafruit_esp32spi_requests as requests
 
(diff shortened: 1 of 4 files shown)


makermelissa:
The Adafruit IO PR  is https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/pull/14 and both should be merged at the same time.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -31,21 +31,20 @@
...
     """
-    def __init__(self, esp, secrets, status_neopixel=None, attempts=2):
+    def __init__(self, esp, secrets, status_pixel=None, attempts=2):
         """
         :param ESP_SPIcontrol esp: The ESP object we are using
         :param dict secrets: The WiFi and Adafruit IO secrets dict (See examples)
         :param int attempts: (Optional) Failed attempts before resetting the ESP32 (default=2)
-        :param status_neopixel: (Optional) The neopixel pin - Usually board.NEOPIXEL (default=None)
-        :type status_neopixel: Pin
+        :param status_pixel: (Optional) The pixel pin - Usually board.NEOPIXEL (default=None)
```
jerryneedell: Why is Pin optional -- why would you pass in a pin name?

makermelissa: Oh, I need to fix that.

makermelissa:
Please note that this PR is very lightly tested as I didn't have all my hardware with me.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
makermelissa: I'll have to wait until I get to my charger though as my laptop with my code died.

makermelissa: Ok, I was able to use the online github editor to fix. The reason it is optional is because the library will function even if you don't pass it in. You just won't be able to visually tell the status.

jerryneedell: np with optional -- I was just confused but the Pin reference -- thanks for fixing. I like this because the user can use any neopixel/dotstar (single or string) available

jerryneedell:
I'll do some testing a bit later today.

brentru:
I've tested this with Adafruit IO CircuitPython and it works without issue on updated examples.

jerryneedell:
has anyone tested it with a dotstar?

makermelissa:
I tested on an ItsyBitsy M4 without an ESP32 attached.

brentru:
Merged in the Adafruit IO CircuitPython PR which referenced this PR, looks good!

PR #16, opened 2019-02-26
Expose the IP Address from WiFiManager

brentru:
Exposes the device's IP address from within WiFiManager. 

Calling `print(wifi.ip_address())` returns `10.0.0.000` (snipped to remove identifiers)

Tested on a PyPortal

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -202,6 +202,16 @@ class ESPSPI_WiFiManager:
         self.pixel_status(0)
         return response_time
 
+    def ip_address(self):
+        """
+        Returns a formatted local IP address, update status pixel.
+        """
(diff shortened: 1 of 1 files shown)


adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -202,6 +202,16 @@ def ping(self, host, ttl=250):
         self.pixel_status(0)
         return response_time
 
+    def ip_address(self):
+        """
+        Returns a formatted local IP address, update status pixel.
+        """
+        if not self._esp.is_connected:
+            self.connect()
```
siddacious: I'm kinda 🤷‍♂️ about things like this while handy do something other than what was requested. I would personally return an error here but this is definitely more convenient.

siddacious: I missed that this is how most of the lib works; no change needed here!

Issue #15, opened 2019-02-26
Expose RSSI in WifiManager for the established connection


siddacious:
It would be great to be able to do something like `wifi.rssi` or more descriptively `wifi.signal_strength` so that one could periodically log/display the signal strength.

brentru:
Great idea - you could expose this method: https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi.py#L386

makermelissa:
That should be easy enough to do.

makermelissa:
Fixed #15 via #16 and #18.

PR #18, opened 2019-02-26
expose rssi from a method in wifimanager

brentru:
Addressing https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/issues/15

Exposing RSSI for the access point we're connected to as `wifi.signal_strength`

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -211,3 +211,11 @@ class ESPSPI_WiFiManager:
         """
         if self.statuspix:
             self.statuspix.fill(value)
+
+    def signal_strength(self):
+        """
+        Returns receiving signal strength indicator in dBm
+        """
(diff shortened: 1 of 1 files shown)


adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -211,3 +211,11 @@ def pixel_status(self, value):
...
         if self.statuspix:
             self.statuspix.fill(value)
+
+    def signal_strength(self):
+        """
+        Returns receiving signal strength indicator in dBm
+        """
+        if not self._esp.is_connected:
+            self.connect()
+        return self._esp.rssi()
```
siddacious: In discord you mentioned that `_esp.rssi()` returns an unpacked `struct` and would need to be cleaned up; is that not the case?

brentru:
I was incorrect on discord, this will return a value (for example, calling `wifi.signal_strength` returns `-47`)

PR #25, opened 2019-03-09
Encode HTTP request header string components as (ASCII) bytes.

anecdata:
Client app code (previously crashed with `TypeError: can't convert str to int` in line 176), e.g.:
```
custom_headers = {"X-Moose" : "Bullwinkle",  "X-Squirrel" : "Rocky", }
response = wifi.get("https://example.com/", data=None, json=None, headers=custom_headers, stream=False)
```
Now results in request headers received by the remote server, e.g.:
```
HTTP_X_MOOSE: Bullwinkle
HTTP_X_SQUIRREL: Rocky
```
(Note: headers are case-insensitive and capitalization varies between server environments.)


ladyada:
not surprising - i never tested that stuff - but @brentru may have!

brentru:
I didn't use custom headers like this originally as I ran into the same issue - was encoding the string from the client's `init`. Don't have to do that anymore though! Tested this PR on a PyPortal with a modified Adafruit IO CircuitPython library - and it's working with examples using post/get.

PR #27, opened 2019-03-12
Fix "Host" and "User-Agent" overrides in request()

cogliano:
In my testing I discovered a "b" is needed in front of the "Host" (line 170) and "User-Agent" (line 172) strings for proper comparison when using request() and sending something like headers={bytes("User-Agent","utf-8"):bytes("Mozilla Clone","utf-8")}


ladyada:
hiya did you check this against the very latest release? we just did some hacking in that area - not sure if this is in conflict or due-to that recent PR

cogliano:
Thank you, Lady Ada, I just downloaded the latest library and it is working without the "b" prefix, so this now works:
request( ..., headers={"User-Agent":"Mozilla Clone"})
I had a site that did not like the default User-Agent and would hang on the request call, now it works like a charm. Thanks.

Issue #28, opened 2019-03-13
requests fails sending large header values or paths over socket


martymcguire:
I'm trying to make an HTTP request with a ~356 byte access token, and when making the request via `requests.get(...)` it fails with:

```
Failed to send 363 bytes (sent 107)
```
With the ESP's `_debug` turned on, I see something like this (redacted for token privacy):

```
...
*** Get host by name
*** Get socket
Allocated socket #0
*** Socket connect mode 2
*** Open socket
Writing: b'GET /microsub/1?action=timeline&channel=duQSjrTyDhvn2BQvrs7jLkV5v4Q4ZYw3 HTTP/1.0\r\n'
Writing: b'Host: aperture.maktro.net\r\n'
Writing: b'User-Agent: Adafruit CircuitPython\r\n'
Writing: b'authorization'
Writing: b': '
Writing: b'Bearer XXXX...FOR_356_BYTES...XXX'
Failed to send 363 bytes (sent 107)
...
```
The server I'm talking to allows me to pass the token in the query string, but I end up with a very similar error (failing on the `GET /microsub/1?...` line).

If I truncate the value of the access token so it's less than ~100 bytes, the request goes through, but the server of course fails it because it's not a valid access token.

It looks like I'm running into an issue with this code: https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi.py#L535

Is it possible to have the requests lib chunk these sends over the socket?

Thanks!!

ladyada:
yep - i think we just haven't sent large chunks of data before, we split into 64 bytes! or you can try to do it at the socket-send level if you want to try :)

kattni:
This was resolved in #29 .

PR #29, opened 2019-03-13
Chunk buffer sends into 64 byte chunks

martymcguire:
Prevents errors with unsent data on large header values, long URL paths

Addresses the specific issue from #28

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -527,11 +527,15 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
         if self._debug:
             print("Writing:", buffer)
         self._socknum_ll[0][0] = socket_num
-        resp = self._send_command_get_response(_SEND_DATA_TCP_CMD,
-                                               (self._socknum_ll[0], buffer),
-                                               sent_param_len_16=True)
+        chunk_size = 64
+        sent = 0
+        for chunk in range((len(buffer) // chunk_size)+1):
+            chunk_buffer = buffer[(chunk*chunk_size):((chunk+1)*chunk_size)]
+            resp = self._send_command_get_response(_SEND_DATA_TCP_CMD,
+                                                   (self._socknum_ll[0], chunk_buffer),
+                                                   sent_param_len_16=True)
+            sent += resp[0][0]
 
-        sent = resp[0][0]
         if sent != len(buffer):
             raise RuntimeError("Failed to send %d bytes (sent %d)" % (len(buffer), sent))
 
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -527,12 +527,11 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
         if self._debug:
             print("Writing:", buffer)
         self._socknum_ll[0][0] = socket_num
-        chunk_size = 64
         sent = 0
-        for chunk in range((len(buffer) // chunk_size)+1):
-            chunk_buffer = buffer[(chunk*chunk_size):((chunk+1)*chunk_size)]
+        for chunk in range((len(buffer) // 64)+1):
             resp = self._send_command_get_response(_SEND_DATA_TCP_CMD,
-                                                   (self._socknum_ll[0], chunk_buffer),
+                                                   (self._socknum_ll[0],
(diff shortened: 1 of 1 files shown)


adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -527,11 +527,15 @@ def socket_write(self, socket_num, buffer):
...
             print("Writing:", buffer)
         self._socknum_ll[0][0] = socket_num
-        resp = self._send_command_get_response(_SEND_DATA_TCP_CMD,
-                                               (self._socknum_ll[0], buffer),
-                                               sent_param_len_16=True)
+        chunk_size = 64
+        sent = 0
+        for chunk in range((len(buffer) // chunk_size)+1):
+            chunk_buffer = buffer[(chunk*chunk_size):((chunk+1)*chunk_size)]
+            resp = self._send_command_get_response(_SEND_DATA_TCP_CMD,
```
jerryneedell: won't this alway send multiples of  64 bytes - even in len(buffer) is not a multiple of 64?
(if len(buffer) < 64 will it send 64?)
If so won't that fail the test at line 539?

martymcguire: thanks for the comment, @jerryneedell ! i was deep into trying to find where the Python docs answer this question before I saw your edit, so I'm pasting this here for posterity:

https://docs.python.org/3/library/stdtypes.html#common-sequence-operations

I find the docs pretty unreadable but my basic understanding is that the end of a slice is a max value, and Python will stop extracting bytes when the buffer being sliced runs out, or when max value is reached.

I was able to confirm by playing on the REPL:

```python
>>> (b'foo bar baz')[0:100]
b'foo bar baz'
```

martymcguire:
update! gonna rewrite this to use MemoryView instead of slices, to avoid all the newly allocated arrays.

martymcguire:
thanks to advice from @ladyada i was able to confirm some memory savings over my first attempt by using `memoryview` instead of buffer slicing and dropping the temp vars.

all numbers come from adding a debug statement into the `for` loop in `socket_write`, and are only a benchmark of my specific demo.

---

```
no memoryview:

Free mem:  145104
Free mem:  143616
Free mem:  142128
Free mem:  140640
Free mem:  139152
Free mem:  137664
Free mem:  136192
Free mem:  147440
Free mem:  147904
Free mem:  147936

memoryview:

Free mem:  145104
Free mem:  143680
Free mem:  142256
Free mem:  140832
Free mem:  139408
Free mem:  137984
Free mem:  136560
Free mem:  147392
Free mem:  147872
Free mem:  147872

memoryview + no chunksize var:

Free mem:  145120
Free mem:  143696
Free mem:  142272
Free mem:  140848
Free mem:  139424
Free mem:  138000
Free mem:  136576
Free mem:  147408
Free mem:  147888
Free mem:  147888

memoryview + no chunksize var + no chunk_buffer var:

Free mem:  145216
Free mem:  143792
Free mem:  142368
Free mem:  140944
Free mem:  139520
Free mem:  138096
Free mem:  136672
Free mem:  147504
Free mem:  147984
Free mem:  147984
```

jerryneedell:
oops - somehow I missed this request for review. I won't be able to do any testing until next week, sorry.  If others have tested it and are comfortable with the change, I have no concern with merging it.

maholli:
Chiming in with a success story.

Using this PR I was able to successfully send large base64 images to adafruit.io that were previously unsuccessful. Much easier than breaking it up into multiple io.send_data calls. 👍 @jerryneedell

maholli:
> 
> 
> Chiming in with a success story.
> 
> Using this PR I was able to successfully send large base64 images to adafruit.io that were previously unsuccessful. Much easier than breaking it up into multiple io.send_data calls. 👍 @jerryneedell

NOTE that JSON errors can still pop up if reading back data from socket (the data was still successfully sent):
```
ESPSocket: 8324 bytes available
Reading 4000 bytes from ESP socket with status 4
ESPSocket: 4324 bytes available
Reading 4000 bytes from ESP socket with status 4
ESPSocket: 324 bytes available
Reading 324 bytes from ESP socket with status 4
ESPSocket: 0 bytes available
syntax error in JSON
```

siddacious:
@ladyada or @martymcguire any reason this shouldn't be merged?

ladyada:
it can be merged

siddacious:
@ladyada done. I can release once the server/AP PRs are merged

ladyada:
that will take a while, you can release now

PR #30, opened 2019-03-15
 fix esp32spi_aio_post.py  -- remove bytes()

jerryneedell:
remove use of bytes() for the header strings


makermelissa:
I’ll test tonight. I don’t have my board with me at the moment.

jerryneedell:
@makermelissa I left it as is for now since the guides that use it tell the user to change it for PyPortal.

ladyada:
thanks folks!

PR #34, opened 2019-04-05
Added support for WPA2 Enterprise, including example code.

docmollo:
This is the CircuiPython code that matches the NINA firmware PR I made yesterday. I've included an example script, as well. I've tested this on a PyPortal and Metro M4, both running CircuitPython 4.0.0-beta.6. I did my best to match the existing style, but I may have not have gotten it right. Let me know if you'd like me to tweak things.


ladyada:
structure looks good to me! once its linted we can merge :)

docmollo:
That was fast....and disapointing.. Looks like it didn't like the line endings?

ladyada:
try running `dos2unix`

docmollo:
One thought I just had... I didin't change the version number in the NINA firmware. I was going to mention something about that, but forgot. The version number in the firmware should probably be rev'd in order to make debugging easier. I could update the example script to contain a message about that new version requirement. And perhaps the README.md could/should contain that requirement.

ladyada:
yep please update the version number too!

PR #38, opened 2019-04-23
Expose GPIO  commands from nina-fw in ESP_SPIcontrol class

cpforbes:
Add set_pin_mode, set_digital_write, and set_analog_write, which simply expose the equivalent functions from nina-fw.

Also added set_esp_debug() to enable debugging output on the ESP32.


brentru:
Looks good! Merging in...

cpforbes:
Thanks.

PR #39, opened 2019-04-25
Updating simpletest/aio_send - adding externally defined pins for breakout, etc

brentru:
Adding externally defined pins to simpletest and aio send. Changing PyPortal to "board with pre-defined ESP32 pins" to support boards like the Metro M4 WiFi.


examples/esp32spi_simpletest.py
```diff
@@ -10,6 +10,19 @@
...
+esp32_reset = DigitalInOut(board.D5)
+
+# If you are using a board with pre-defined ESP32 Pins:
+"""
+esp32_cs = DigitalInOut(board.ESP_CS)
+esp32_ready = DigitalInOut(board.ESP_BUSY)
+esp32_reset = DigitalInOut(board.ESP_RESET)
+"""
+
 esp32_cs = DigitalInOut(board.ESP_CS)
```
jerryneedell: these override the lines above (lines 15-17)

jerryneedell: I think you meant to remove lines 26-28

brentru: Yup, adding in on #952b38b, thanks for the catch!

brentru:
@kattni  Ok, updated to use `#` comments.

PR #43, opened 2019-05-10
Fix zero timeout in socket, add configurable timeout to ESP_SPIcontrol

cpforbes:
- This fixes a bug where a zero timeout in socket would not block as documented.

- Adds a configurable timeout to ESP_SPIcontrol.  Some SSL sites took longer than 10 seconds to complete the SSL handshake and the communication with the ESP32 timed out.  Increasing the timeout to 12 seconds resolves the problem described in https://github.com/arduino/nina-fw/issues/10

- Handle missing optional conntype parameter in socket.connect.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -126,14 +126,25 @@ WL_AP_FAILED          = const(9)
 
 class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
     """A class that will talk to an ESP32 module programmed with special firmware
-    that lets it act as a fast an efficient WiFi co-processor"""
+    that lets it act as a fast an efficient WiFi co-processor
+
+    :param busio.SPI spi_bus: The SPI bus the ESP32 is connected to.
+    :param digitalio.DigitalInOut cs_pin: The SPI CS pin.
+    :param digitalio.DigitalInOut ready_pin: The ready pin.
+    :param digitalio.DigitalInOut reset_pin: The reset pin.
+    :param digitalio.DigitalInOut gpio0_pin: (optional) Pin connected to GPIO 0.
+    :param int timeout: (optional) Command response timeout duration in seconds (default = 10).
+    :param int debug: (optional) Enable debug output (3 = maximum debug)
+    """
     TCP_MODE = const(0)
     UDP_MODE = const(1)
     TLS_MODE = const(2)
 
     # pylint: disable=too-many-arguments
-    def __init__(self, spi, cs_pin, ready_pin, reset_pin, gpio0_pin=None, *, debug=False):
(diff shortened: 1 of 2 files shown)


adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -126,14 +126,25 @@
...
+    :param int debug: (optional) Enable debug output (3 = maximum debug)
+    """
     TCP_MODE = const(0)
     UDP_MODE = const(1)
     TLS_MODE = const(2)
 
     # pylint: disable=too-many-arguments
-    def __init__(self, spi, cs_pin, ready_pin, reset_pin, gpio0_pin=None, *, debug=False):
+    def __init__(self, spi, cs_pin, ready_pin, reset_pin, gpio0_pin=None, *,
+                 timeout=10, debug=False):
```
ladyada: you want the timeout to be 12 or 15?

ladyada: (FYI https://github.com/arduino/nina-fw/issues/10 was not related to timeout, it was fixed with some SSL reconfiguration)

cpforbes: I don't see much harm in increasing the default timeout since it only impacts waiting for the response from the ESP which (hopefully) should rarely time out. I didn't change it in this PR so the default behavior would not change.

I can update the PR with a longer default if you would like.

Interesting about that bug.  I found that that site consistently took longer than 10 seconds and timed out for me.

ladyada: ok i dont know that this PR does what you think? the timetime here is for the ESP to communicate over SPI, has nothing to do with sockets :?

cpforbes: This PR confusingly has 2 timeout changes -- I probably should have split them up.

In the socket class there was a bug where a timeout of zero did not block as documented in socket.settimeout().

I discovered that while testing the configurable ESP32 SPI command timeout.

The configurable SPI timeout is intended.  I was curious about https://github.com/arduino/nina-fw/issues/10 and tried to replicate it. The debug output on the ESP32 showed it completing a few seconds after circuitpython timed out waiting for the command response. (I had a separate FTDI connected to the ESP32 to monitor it's output)

ladyada: how about we remove the SPI timeout stuff and just do the socket fix in this PR - cause im still not sure about the SPI timeout, there shouldnt be such a long timeout required for SPI :/

cpforbes: Ok I will close this and submit a new PR with just the socket timeout fix.

I will collect more data on the SPI timeout and open an issue if needed.

ladyada: thanks :)

PR #45, opened 2019-05-19
Adding WPA2 Enterprise support to the WiFi manager library

docmollo:
This is my first attempt to add support for WPA2 Enterprise to the WiFi Manager library. It all works in my testing environment. The one thing I was thinking about, but didn't add, was support to check the NINA firmware version and bail out if the ESP32 isn't running a compatible version.

NOTE: Pylint returns some not-happy results, but I'm not sure if they are going to be issues. If so, I'm not sure how to go about resolving them. Some guideance would be appreciated.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -31,25 +31,18 @@ WiFi Manager for making ESP32 SPI as WiFi much easier
 
 # pylint: disable=no-name-in-module
 
+from micropython import const
 from adafruit_esp32spi import adafruit_esp32spi
 import adafruit_esp32spi.adafruit_esp32spi_requests as requests
-
-class WiFiConnType: # pylint: disable=too-few-public-methods
-    """An enum-like class representing the different types of WiFi connections
-    that can be made. The values can be referenced like ``WiFiConnType.normal``.
-    Possible values are
-    - ``ThermocoupleType.normal``
-    - ``ThermocoupleType.enterprise``
-    """
-    # pylint: disable=invalid-name
-    normal = 1
-    enterprise = 2
 
 class ESPSPI_WiFiManager:
     """
     A class to help manage the Wifi connection
     """
-    def __init__(self, esp, secrets, status_pixel=None, attempts=2, wificonntype=WiFiConnType.normal):
+    NORMAL = const(1)
+    ENTERPRISE = const(2)
+
+    def __init__(self, esp, secrets, status_pixel=None, attempts=2, connection_type=NORMAL):
         """
         :param ESP_SPIcontrol esp: The ESP object we are using
         :param dict secrets: The WiFi and Adafruit IO secrets dict (See examples)
@@ -69,7 +62,7 @@ class ESPSPI_WiFiManager:
         self.ent_ssid = secrets['ent_ssid']
         self.ent_ident = secrets['ent_ident']
         self.ent_user = secrets['ent_user']
-        self.ent_passwd = secrets['ent_passwd']
+        self.ent_password = secrets['ent_password']
         self.attempts = attempts
         self._connection_type = connection_type
         requests.set_interface(self.esp)
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -36,6 +36,17 @@ from micropython import const
 from adafruit_esp32spi import adafruit_esp32spi
 import adafruit_esp32spi.adafruit_esp32spi_requests as requests
 
+class WiFiConnType: # pylint: disable=too-few-public-methods
+    """An enum-like class representing the different types of WiFi connections
+    that can be made. The values can be referenced like ``WiFiConnType.normal``.
+    Possible values are
+    - ``ThermocoupleType.normal``
+    - ``ThermocoupleType.enterprise``
+    """
+    # pylint: disable=invalid-name
+    normal = 1
+    enterprise = 2
+
 class ESPSPI_WiFiManager:
     """
     A class to help manage the Wifi connection
@@ -117,12 +128,12 @@ class ESPSPI_WiFiManager:
         Attempt a regular style WiFi connection
         """
         failure_count = 0
-        while not self._esp.is_connected:
+        while not self.esp.is_connected:
             try:
                 if self.debug:
                     print("Connecting to AP...")
                 self.pixel_status((100, 0, 0))
-                self._esp.connect_AP(bytes(self.ssid, 'utf-8'), bytes(self.password, 'utf-8'))
+                self.esp.connect_AP(bytes(self.ssid, 'utf-8'), bytes(self.password, 'utf-8'))
                 failure_count = 0
                 self.pixel_status((0, 100, 0))
             except (ValueError, RuntimeError) as error:
@@ -138,12 +149,12 @@ class ESPSPI_WiFiManager:
         Attempt an enterprise style WiFi connection
         """
         failure_count = 0
-        self._esp.wifi_set_network(bytes(self.ent_ssid, 'utf-8'))
-        self._esp.wifi_set_entidentity(bytes(self.ent_ident, 'utf-8'))
-        self._esp.wifi_set_entusername(bytes(self.ent_user, 'utf-8'))
-        self._esp.wifi_set_entpassword(bytes(self.ent_password, 'utf-8'))
-        self._esp.wifi_set_entenable()
-        while not self._esp.is_connected:
+        self.esp.wifi_set_network(bytes(self.ent_ssid, 'utf-8'))
+        self.esp.wifi_set_entidentity(bytes(self.ent_ident, 'utf-8'))
+        self.esp.wifi_set_entusername(bytes(self.ent_user, 'utf-8'))
+        self.esp.wifi_set_entpassword(bytes(self.ent_password, 'utf-8'))
+        self.esp.wifi_set_entenable()
+        while not self.esp.is_connected:
             try:
                 if self.debug:
                     print("Waiting for the ESP32 to connect to the WPA2 Enterprise AP...")
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -57,17 +57,13 @@ class ESPSPI_WiFiManager:
         connection to make. The default is "normal".
         """
         # Read the settings
-        self.esp = esp
+        self._esp = esp
         self.debug = False
         self.ssid = secrets['ssid']
         self.password = secrets['password']
-        self.ent_ssid = secrets['ent_ssid']
-        self.ent_ident = secrets['ent_ident']
-        self.ent_user = secrets['ent_user']
-        self.ent_password = secrets['ent_password']
         self.attempts = attempts
         self._connection_type = connection_type
-        requests.set_interface(self.esp)
+        requests.set_interface(self._esp)
         self.statuspix = status_pixel
         self.pixel_status(0)
 
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -52,18 +52,16 @@ class ESPSPI_WiFiManager:
             or RGB LED (default=None)
         :type status_pixel: NeoPixel, DotStar, or RGB LED
         :param int attempts: (Optional) Failed attempts before resetting the ESP32 (default=2)
-        :param const con_type: (Optional) Type of WiFi connection: normal=1, WPA2 Enterprise=2
-        :param ~adafruit_esp32spi_wifimanager.WiFiConnType wificonntype: The type of WiFi \
-        connection to make. The default is "normal".
+        :param const connection_type: (Optional) Type of WiFi connection: NORMAL or ENTERPRISE
         """
         # Read the settings
-        self._esp = esp
+        self.esp = esp
         self.debug = False
         self.ssid = secrets['ssid']
         self.password = secrets['password']
         self.attempts = attempts
         self._connection_type = connection_type
-        requests.set_interface(self._esp)
+        requests.set_interface(self.esp)
         self.statuspix = status_pixel
         self.pixel_status(0)
 
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -38,7 +38,7 @@ class ESPSPI_WiFiManager:
     """
     A class to help manage the Wifi connection
     """
-    def __init__(self, esp, secrets, status_pixel=None, attempts=2):
+    def __init__(self, esp, secrets, status_pixel=None, attempts=2, con_type=1):
         """
         :param ESP_SPIcontrol esp: The ESP object we are using
         :param dict secrets: The WiFi and Adafruit IO secrets dict (See examples)
@@ -46,14 +46,20 @@ class ESPSPI_WiFiManager:
             or RGB LED (default=None)
         :type status_pixel: NeoPixel, DotStar, or RGB LED
         :param int attempts: (Optional) Failed attempts before resetting the ESP32 (default=2)
+        :param int con_type: (Optional) Type of WiFi connection to make: normal=1, WPA2 Enterprise=2
         """
         # Read the settings
         self.esp = esp
         self.debug = False
         self.ssid = secrets['ssid']
         self.password = secrets['password']
+        self.ent_ssid = secrets['ent_ssid']
+        self.ent_ident = secrets['ent_ident']
+        self.ent_user = secrets['ent_user']
+        self.ent_passwd = secrets['ent_passwd']
         self.attempts = attempts
         requests.set_interface(self.esp)
+        self.con_type = con_type
         self.statuspix = status_pixel
         self.pixel_status(0)
 
@@ -76,6 +82,17 @@ class ESPSPI_WiFiManager:
             print("MAC addr:", [hex(i) for i in self.esp.MAC_address])
             for access_pt in self.esp.scan_networks():
                 print("\t%s\t\tRSSI: %d" % (str(access_pt['ssid'], 'utf-8'), access_pt['rssi']))
+        if self._connection_type == ESPSPI_WiFiManager.NORMAL:
+            self.connect_normal()
+        elif self._connection_type == ESPSPI_WiFiManager.ENTERPRISE:
+            self.connect_enterprise()
+        else:
+            raise TypeError("Invalid WiFi connection type specified")
+
+    def connect_normal(self):
+        """
+        Attempt a regular style WiFi connection
+        """
         failure_count = 0
         while not self.esp.is_connected:
             try:
@@ -85,6 +102,33 @@ class ESPSPI_WiFiManager:
                 self.esp.connect_AP(bytes(self.ssid, 'utf-8'), bytes(self.password, 'utf-8'))
                 failure_count = 0
                 self.pixel_status((0, 100, 0))
+            except (ValueError, RuntimeError) as error:
+                print("Failed to connect, retrying\n", error)
+                failure_count += 1
+                if failure_count >= self.attempts:
+                    failure_count = 0
+                    self.reset()
+                continue
+
+    def connect_enterprise(self):
+        """
+        Attempt an enterprise style WiFi connection
+        """
+        failure_count = 0
+        self.esp.wifi_set_network(bytes(self.ent_ssid, 'utf-8'))
+        self.esp.wifi_set_entidentity(bytes(self.ent_ident, 'utf-8'))
+        self.esp.wifi_set_entusername(bytes(self.ent_user, 'utf-8'))
+        self.esp.wifi_set_entpassword(bytes(self.ent_password, 'utf-8'))
+        self.esp.wifi_set_entenable()
+        while not self.esp.is_connected:
+            try:
+                if self.debug:
+                    print("Waiting for the ESP32 to connect to the WPA2 Enterprise AP...")
+                self.pixel_status((100, 0, 0))
+                sleep(1)
+                failure_count = 0
+                self.pixel_status((0, 100, 0))
+                sleep(1)
             except (ValueError, RuntimeError) as error:
                 print("Failed to connect, retrying\n", error)
                 failure_count += 1
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -48,12 +48,12 @@ class ESPSPI_WiFiManager:
         :param int attempts: (Optional) Failed attempts before resetting the ESP32 (default=2)
         """
         # Read the settings
-        self._esp = esp
+        self.esp = esp
         self.debug = False
         self.ssid = secrets['ssid']
         self.password = secrets['password']
         self.attempts = attempts
-        requests.set_interface(self._esp)
+        requests.set_interface(self.esp)
         self.statuspix = status_pixel
         self.pixel_status(0)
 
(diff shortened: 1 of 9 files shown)


ladyada:
nice, please make `:param int con_type` an enum or at least a named constant!

docmollo:
So...I thought I could just google this and figure it out, but this is getting into actual CS stuff that I haven't dealt with before. Is the fix as simple as just changing the "param int" to "param enum"?

ladyada:
ok make them named constants then!

docmollo:
Ah ha! I think I figured out what you're asking for. Are you suggesting something like this class in the MAX31856 library?

https://github.com/adafruit/Adafruit_CircuitPython_MAX31856/blob/master/adafruit_max31856.py#L95

ladyada:
yes!

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -34,24 +34,43 @@
...
 
+class WiFiConnType: # pylint: disable=too-few-public-methods
+    """An enum-like class representing the different types of WiFi connections
+    that can be made. The values can be referenced like ``WiFiConnType.normal``.
+    Possible values are
+    - ``ThermocoupleType.normal``
+    - ``ThermocoupleType.enterprise``
+    """
+    # pylint: disable=invalid-name
+    normal = 1
```
dhalbert: Make `normal` and `enterprise` be all caps, since these are constants. Then you can take away `# pylint: disable=invalid-name`.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -34,24 +34,43 @@
...
+    enterprise = 2
+
 class ESPSPI_WiFiManager:
     """
     A class to help manage the Wifi connection
     """
-    def __init__(self, esp, secrets, status_pixel=None, attempts=2):
+    def __init__(self, esp, secrets, status_pixel=None, attempts=2, wificonntype=WiFiConnType.normal):
```
dhalbert: I'd use `connection_type` instead of `wificonntype`. We know it's wifi, and normal practice to to separate words with underscores.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -34,24 +34,43 @@
 from adafruit_esp32spi import adafruit_esp32spi
 import adafruit_esp32spi.adafruit_esp32spi_requests as requests
 
+class WiFiConnType: # pylint: disable=too-few-public-methods
```
dhalbert: Instead of using a separate class, you could just add these constants to `ESPSPI_WiFiManager`.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -34,18 +34,31 @@
...
+    """An enum-like class representing the different types of WiFi connections
+    that can be made. The values can be referenced like ``WiFiConnType.normal``.
+    Possible values are
+    - ``ThermocoupleType.normal``
```
brentru: Change to `WiFiConType.normal`

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -34,18 +34,31 @@
...
+    - ``ThermocoupleType.enterprise``
```
brentru: Change to `WiFiConType.enterprise`

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -34,24 +34,43 @@
...
+        :param const con_type: (Optional) Type of WiFi connection: normal=1, WPA2 Enterprise=2
+        :param ~adafruit_esp32spi_wifimanager.WiFiConnType wificonntype: The type of WiFi \
+        connection to make. The default is "normal".
         """
         # Read the settings
         self._esp = esp
         self.debug = False
         self.ssid = secrets['ssid']
         self.password = secrets['password']
+        self.ent_ssid = secrets['ent_ssid']
```
brentru: I feel using the regular `secrets['ssid']` is OK here.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -34,24 +34,43 @@
...
+        self.ent_ident = secrets['ent_ident']
+        self.ent_user = secrets['ent_user']
+        self.ent_passwd = secrets['ent_passwd']
```
brentru: Change this to `secrets['ent_password']` to keep convention with `secrets['password']`

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -34,24 +34,43 @@
...
         self.attempts = attempts
+        self.wificonntype = wificonntype
```
brentru: Switch to `self._wificonntype` instead. This is only used within `connect()`. I'm not certain its a `wifi` property the user would want to modify.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -76,21 +95,45 @@ def connect(self):
...
+                    self.pixel_status((0, 100, 0))
+                except (ValueError, RuntimeError) as error:
+                    print("Failed to connect, retrying\n", error)
+                    failure_count += 1
+                    if failure_count >= self.attempts:
+                        failure_count = 0
+                        self.reset()
+                    continue
+        elif self.wificonntype == WiFiConnType.enterprise:
+            self._esp.wifi_set_network(bytes(self.ent_ssid, 'utf-8'))
```
brentru: change to `ssid`

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -76,21 +95,45 @@ def connect(self):
...
+                    self.pixel_status((0, 100, 0))
+                except (ValueError, RuntimeError) as error:
+                    print("Failed to connect, retrying\n", error)
+                    failure_count += 1
+                    if failure_count >= self.attempts:
+                        failure_count = 0
+                        self.reset()
+                    continue
+        else:
+            print("Invalid connection type! Exiting...")
```
brentru: CircuitPython libraries don't normally `print` and `exit`. They usually `raise` an error. Instead of L135/L136,

`raise TypeError('Invalid wificonntype specified.')`

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: Agreed. Fixed as part of one of your other suggestions.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: Agreed. Fixed in a coming commit

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: Agreed. After I put the class together, I thought to myself that this seemed like overkill for 2 options.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: This class is being removed in an upcoming commit and replaced with a couple of constants.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: This class is being removed in an upcoming commit and replaced with a couple of constants.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: I thought about that, but I like having one secrets.py file with all of the options in it. At the very least, it makes testing easier, since I'm switching back and forth testing both regular and enterprise connections during development. I'm not going to get to it tonight, but what do you think about adding some code that would support both, but fall back to the regulat ssid if the ent_ssid wasn't in the secrets.py file?

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: Agreed. Fixed in an upcoming commit.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: Agreed. Per one of @dhalbert's suggestions, this is now self._connection_type in an upcoming commit.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: This is being discussed in another comment/suggestion. Resolving here and will address as needed.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: I was thinking there has to be a better way/more Python way of dealing with this. And also realized I never tested this code. Fixed in an upcoming commit.

docmollo:
Just a heads-up - I took @kattni up on her offer to assist with the pylint issues. We're scheduled to chat about this on Thursday (5/23) evening, so there are other commits coming.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
brentru: Detecting the `ent_ssid` and falling back to `ssid` would be a great solution. My main issue with the way that this was done was the creation of more variables (possibly unused/broken, wouldnt find `secrets['ssid']` at all).

examples/esp32spi_wpa2ent_aio_post.py
```diff
@@ -0,0 +1,57 @@
...
+print("ESP32 SPI WPA2 Enterprise webclient test")
+
+# Get wifi details and more from a secrets.py file
+try:
+    from secrets import secrets
+except ImportError:
+    print("WiFi secrets are kept in secrets.py, please add them there!")
+    raise
+
+# If you are using a board with pre-defined ESP32 Pins:
```
brentru: Check out how we automatically switch between the ESP32 pin definitions in CircuitPython_IO examples: https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/blob/master/examples/adafruit_io_simpletest_data.py#L25

brentru: Check out how we automatically switch between the ESP32 pin definitions in CircuitPython_IO examples: https://github.com/adafruit/Adafruit_CircuitPython_AdafruitIO/blob/master/examples/adafruit_io_simpletest_data.py#L25

docmollo:
Had my meeting with @kattni this evening. I'm going to refactor the `connect()` function into two different functions for the two different types of connections. As part of that work, I'm also going to work on the ideas @brentru and I have been discussing. It's probably going to take me a week or two to get to all the changes complete, so if you would like a status, feel free to ping me here or on Discord.

examples/esp32spi_wpa2ent_aio_post.py
docmollo: Oh, nice! I'm going to add that change to my next commit. Thanks for the tip!

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: OK...I took a stab at this in my latest code. Commit coming soon!

docmollo:
OK...sorry for the huge delay! I have finally coded up all the changes I've talked about with @kattni and @brentru. Let me know if there is anything else

jerryneedell:
@docmollo I'm planning  to do some testing with this soon.  If not using WPA2 Enterprise, does anything in they PR require your version of the Nina FW. If so, what version is required --
The guide page has 1.2.2 and 1.3.0  -- should this all (except for Enterprise ) work on both? https://learn.adafruit.com/adding-a-wifi-co-processor-to-circuitpython-esp8266-esp32/firmware-files
also -- is the current master for the Adafruit nina-fw repo  https://github.com/adafruit/nina-fw 
up to date with your mods so a build from there should work, correct?

docmollo:
@jerryneedell in theory, the WPA2 Enterprise support in the CPy library should *not* affect non-WPA2 Enterprise functionality. It's possible that prior to this PR, if you didn't have the related settings in the secrets.py file, the code might have barfed....that was one of the tweaks in this PR - to make that regular WiFi functionality work w/o having to possibly jump through unnecessary hoops. Let me know if you run into issues.

Even though you can't test it, you mentioned it, so here's the answer to your nina-fw question. The WPA2 Enterprise CPy code requires the nina-fw to be 1.3.0 or newer (greater?). I don't remember which example off the top of my head as I type this, but one of the two examples I have has an example of checking the firmware version number to make sure it is what it needs.

jerryneedell:
Ran into an issue loading the latest nina-fw  file:
Loading the  NINA_W102-1.3.1.bin  file via the Metro_m4 passthru on an metro_m4_airlift.  If I build it from your master repo, it loads and works OK but if I buil it from the Adafruit nina_fw repo, the load fails partway trough with a timeout ... see examples

this fails
``` 
jerryneedell@Ubuntu-Macmini:~/projects/adafruit_github/nina-fw$ /usr/local/bin/esptool.py  --port /dev/ttyACM0 --before no_reset --baud 115200 write_flash 0 NINA_W102-1.3.1.bin 
esptool.py v2.5.1
Serial port /dev/ttyACM0
Connecting.....
Detecting chip type... ESP32
Chip is ESP32D0WDQ6 (revision 1)
Features: WiFi, BT, Dual Core, 240MHz, VRef calibration in efuse
MAC: c4:4f:33:0e:a9:dd
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Auto-detected Flash size: 4MB
Compressed 1129472 bytes to 610164...
Writing at 0x00024000... (26 %)
A fatal error occurred: Timed out waiting for packet header
```



this works
```
jerryneedell@Ubuntu-Macmini:~/projects/adafruit_github/nina-fw$ cd ../../docmollo/nina-fw/
jerryneedell@Ubuntu-Macmini:~/projects/docmollo/nina-fw$ /usr/local/bin/esptool.py  --port /dev/ttyACM0 --before no_reset --baud 115200 write_flash 0 NINA_W102-1.3.1.bin 
esptool.py v2.5.1
Serial port /dev/ttyACM0
Connecting.....
Detecting chip type... ESP32
Chip is ESP32D0WDQ6 (revision 1)
Features: WiFi, BT, Dual Core, 240MHz, VRef calibration in efuse
MAC: c4:4f:33:0e:a9:dd
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Auto-detected Flash size: 4MB
Compressed 1129472 bytes to 610173...
Wrote 1129472 bytes (610173 compressed) at 0x00000000 in 53.7 seconds (effective 168.4 kbit/s)...
Hash of data verified.

Leaving...
Hard resetting via RTS pin...
  first is from Adfruit repo, second from yours  -- for both I just do make the 'make firmware`  Have you built it from the Adafruit repo?
This is repeatable...
```

I was able to work around it with the help of @NeverCast in discord
 
The adafruit repo build loads OK with:
```
jerryneedell@Ubuntu-Macmini:~/projects/adafruit_github/nina-fw$ /usr/local/bin/esptool.py --chip esp32 --port /dev/ttyACM0 --baud 115200 --before no_reset  write_flash -z --flash_mode dio --flash_freq 40m --flash_size detect 0x1000 /home/jerryneedell/projects/adafruit_github/nina-fw/build/bootloader/bootloader.bin 0x30000 /home/jerryneedell/projects/adafruit_github/nina-fw/build/nina-fw.bin 0x8000 /home/jerryneedell/projects/adafruit_github/nina-fw/build/partitions.bin
esptool.py v2.5.1
Serial port /dev/ttyACM0
Connecting....
Chip is ESP32D0WDQ6 (revision 1)
Features: WiFi, BT, Dual Core, 240MHz, VRef calibration in efuse
MAC: c4:4f:33:0e:a9:dd
Uploading stub...
Running stub...
Stub running...
Configuring flash size...
Auto-detected Flash size: 4MB
Flash params set to 0x0220
Compressed 15168 bytes to 9930...
Wrote 15168 bytes (9930 compressed) at 0x00001000 in 0.9 seconds (effective 139.3 kbit/s)...
Hash of data verified.
Compressed 932672 bytes to 564789...
Wrote 932672 bytes (564789 compressed) at 0x00030000 in 49.6 seconds (effective 150.3 kbit/s)...
Hash of data verified.
Compressed 3072 bytes to 120...
Wrote 3072 bytes (120 compressed) at 0x00008000 in 0.0 seconds (effective 1733.1 kbit/s)...
Hash of data verified.

Leaving...
Hard resetting via RTS pin...
```

I still do not understand what is different between the builds but it is working now...

nevercast:
I suspected that the padding used in the flat firmware binary was causing a brown out, so I suggested flashing the individual partitions seperately. This seems to have worked.

It seems the partition for the application is at 0x30000 which means there is a lot of padding between the partition table at 0x8000 and the application code.

It may resolve the situation if the application can instead be flashes at 0x10000, but I suspect that would mask the problem, the real issue may be in hardware if the ESP32 really is browning out.

I don't own an M4 Airlift so I cannot test this on equipment here.

jerryneedell:
ok -- now relevant to this PR - I tried my often used "cherlights" test with this PR and there appears to have been some breaking change -- this code worked on the previous release
```
Press any key to enter the REPL. Use CTRL-D to reload.
Adafruit CircuitPython 4.1.0-beta.0 on 2019-06-14; Adafruit Metro M4 Airlift Lite with samd51j19
>>> 
>>> import cheer_ring
ESP32 SPI webclient test
Fetching json from https://api.thingspeak.com/channels/1417/feeds.json?results=1
Failed to connect, retrying
 ('No such ssid', b'Needell Airport')
  File "<stdin>", line 1, in <module>
  File "cheer_ring.py", line 55, in <module>
  File "cheer_ring.py", line 45, in <module>
  File "adafruit_esp32spi/adafruit_esp32spi_wifimanager.py", line 170, in get
  File "adafruit_esp32spi/adafruit_esp32spi_wifimanager.py", line 105, in connect
  File "adafruit_esp32spi/adafruit_esp32spi_wifimanager.py", line 129, in connect_normal
  File "adafruit_esp32spi/adafruit_esp32spi_wifimanager.py", line 125, in connect_normal
NameError: local variable referenced before assignment
>>> 
```

code
```
import time
import board
import busio
from digitalio import DigitalInOut

from adafruit_esp32spi import adafruit_esp32spi_wifimanager

import neopixel
import adafruit_fancyled.adafruit_fancyled as fancy

# Get wifi details and more from a secrets.py file
try:
    from secrets import secrets
except ImportError:
    print("WiFi secrets are kept in secrets.py, please add them there!")
    raise

print("ESP32 SPI webclient test")

DATA_SOURCE = "https://api.thingspeak.com/channels/1417/feeds.json?results=1"
DATA_LOCATION = ["feeds", 0, "field2"]

esp32_cs = DigitalInOut(board.ESP_CS)
esp32_ready = DigitalInOut(board.ESP_BUSY)
esp32_reset = DigitalInOut(board.ESP_RESET)
spi = busio.SPI(board.SCK, board.MOSI, board.MISO)
esp = adafruit_esp32spi.ESP_SPIcontrol(spi, esp32_cs, esp32_ready, esp32_reset)
"""Use below for Most Boards"""
status_light = neopixel.NeoPixel(board.NEOPIXEL, 1, brightness=0.2) # Uncomment for Most Boards
"""Uncomment below for ItsyBitsy M4"""
#status_light = dotstar.DotStar(board.APA102_SCK, board.APA102_MOSI, 1, brightness=0.2)
wifi = adafruit_esp32spi_wifimanager.ESPSPI_WiFiManager(esp, secrets, status_light)

# neopixels
pixels = neopixel.NeoPixel(board.A1, 16, brightness=0.3)
pixels.fill(0)

# we'll save the value in question
last_value = value = None

while True:
    try:
        print("Fetching json from", DATA_SOURCE)
        response = wifi.get(DATA_SOURCE)
        print(response.json())
        value = response.json()
        for key in DATA_LOCATION:
            value = value[key]
            print(value)
        response.close()
    except (ValueError, RuntimeError) as e:
        print("Failed to get data, retrying\n", e)
        wifi.reset()
        continue

    if not value:
        continue
    if last_value != value:
        color = int(value[1:], 16)
        red = color >> 16 & 0xFF
        green = color >> 8 & 0xFF
        blue = color& 0xFF
        gamma_corrected = fancy.gamma_adjust(fancy.CRGB(red, green, blue)).pack()

        pixels.fill(gamma_corrected)
```

jerryneedell:
hmmm -- on a second run -- it ran ok  -- looking further into it.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -75,7 +101,17 @@ def connect(self):
             print("MAC addr:", [hex(i) for i in self._esp.MAC_address])
             for access_pt in self._esp.scan_networks():
                 print("\t%s\t\tRSSI: %d" % (str(access_pt['ssid'], 'utf-8'), access_pt['rssi']))
-        failure_count = 0
```
jerryneedell: Removing this line breaks line 125 below

jerryneedell:
as imple fix -- just have to move the line taht initializes failure_cunt up before the attempt to connect or esle it throws the exception before it gets initialized
```
diff --git a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
index 6f7c944..0ba847c 100755
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -117,8 +117,8 @@ class ESPSPI_WiFiManager:
                 if self.debug:
                     print("Connecting to AP...")
                 self.pixel_status((100, 0, 0))
-                self._esp.connect_AP(bytes(self.ssid, 'utf-8'), bytes(self.password, 'utf-8'))
                 failure_count = 0
+                self._esp.connect_AP(bytes(self.ssid, 'utf-8'), bytes(self.password, 'utf-8'))
                 self.pixel_status((0, 100, 0))
             except (ValueError, RuntimeError) as error:
                 print("Failed to connect, retrying\n", error)
```

jerryneedell:
oops  - the proposed fix above is not correct -- it will circumvent the maximum number of retries.
failure_count must be initialized to 0 before the try/except loop stars -- it also should be reset to 0 after a successful connection.
Sorry for the confusion.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -92,6 +128,32 @@ def connect(self):
...
+        self._esp.wifi_set_entusername(bytes(self.ent_user, 'utf-8'))
+        self._esp.wifi_set_entpassword(bytes(self.ent_password, 'utf-8'))
+        self._esp.wifi_set_entenable()
+        while not self._esp.is_connected:
+            try:
+                if self.debug:
+                    print("Waiting for the ESP32 to connect to the WPA2 Enterprise AP...")
+                self.pixel_status((100, 0, 0))
+                sleep(1)
+                failure_count = 0
```
jerryneedell: This is a bit different form the "normal" connect since the actual connect is not in the try/except loop. It's not clear to me how this works...I looks like it just waits for a connection but as constructed, I don't think it will ever time out and it is not clear that the failure_count test is meaningful. 
An exception in lines 135-139 with not be "handled" and will simply raise the exception. Is that the desired response?

Or I am just confused...

docmollo:
Thanks @jerryneedell and @nevercast for poking at the nina-fw, but I'm going to ignore those issues mentioned above since they are unrelated to this PR. I'll follow-up on the issue opened on the nina-fw repo when I get a chance (probably later this week or weekend).

With regards to the failure_count issue, yeah, that's my fault. As soon as you mentioned it, I knew what had happened. Work suddenly got very busy yesterday after the CP weekly meeting, so I'm only going to have time after work this week to work on fixing the issue. Worst case, I'll definitely have time this weekend.

Thanks again!

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: Fixed in the latest commit.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
docmollo: Fixed in the latest commit.

docmollo:
@jerryneedell if you could give this fix a try, I'd appreciate it

docmollo:
OK....I think this will do it, @kattni. Much hacking of code and poking at git, and I think I have it. Let me know if you catch something I missed....totally possible at this point ;)

Issue #47, opened 2019-05-28
Expose RGB LEDs in WiFiManager, instead of NeoPixels


brentru:
Some of the AirLift breakouts use RGB LEDS instead of NeoPixels. It'd be nice to have an status indicator LED within the WiFiManager for cases when NeoPixels aren't present.

The AirLift Breakout follows the pin ordering: 25 (Red), 26 (Green), 27 (Blue).

makermelissa:
Do you know if we have a library to set the color of RGB leds? I had abstracted it so either a DotStar or Neopixel could be passed in. Probably not, but I’m just hoping it will be that easy to implement. :)

brentru:
@makermelissa  Yup! The LED pins can be defined in ESP32SPI (https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi.py#L629)

and written to as outputs, in `set_digital_write` (https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi.py#L647).

You can also PWM the RGB LEDs:
https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi.py#L659

Since it's the same across the airlift hardware, one could add a toggle for a RGB status light. There'll need to be a helper-RGB-LED method within WiFiManager too, to do some color-mixin'.

Here's a really basic example of setting and un-setting each individual RGB LED which works with _all_ AirLift hardware with the RGB LEDs. 
```
spi = busio.SPI(board.SCK, board.MOSI, board.MISO)
esp = adafruit_esp32spi.ESP_SPIcontrol(spi, esp32_cs, esp32_ready, esp32_reset)
import time
pins = [25, 26, 27]

for i in pins:
    esp.set_pin_mode(i, 1)
    esp.set_digital_write(i, 1)
    time.sleep(1)
    esp.set_digital_write(i, 0)
```

makermelissa:
It might be good to create another class in this library for RGB leds that handles the color mixing that can be passed in like the DotStar or Neopixel.

brentru:
Agreed!

anecdata:
[...]

PR #53, opened 2019-06-11
Add RGB Status LED to WiFiManager

brentru:
This pull request adds the option of using an externally defined RGB LED (https://github.com/adafruit/Adafruit_CircuitPython_RGBLED) as a `status_pixel` for `WiFiManager`. Useful for AirLift Breakout and FeatherWing.

* `pixel_status` has been modified to set `color` based on LED type.
*  Modified `examples/esp32spi_aio_post.py` to show usage for a RGB LED.

Tested with Feather M4 on CircuitPython 4.0.1, AirLift FeatherWing. Tested both constructing and passing `wifi` an RGB LED `status_pixel` and a NeoPixel `status_pixel`. 

Addressing https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/issues/47

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
@@ -55,9 +55,6 @@ class ESPSPI_WiFiManager:
         self.attempts = attempts
         requests.set_interface(self._esp)
         self.statuspix = status_pixel
-        self._is_rgb_led = False
-        if hasattr(self.statuspix, 'color'):
-            self._is_rgb_led = True
         self.pixel_status(0)
 
     def reset(self):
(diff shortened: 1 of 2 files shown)


adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -54,6 +55,9 @@ def __init__(self, esp, secrets, status_pixel=None, attempts=2):
         self.attempts = attempts
         requests.set_interface(self._esp)
         self.statuspix = status_pixel
+        self._is_rgb_led = False
```
ladyada: there is no need to have an additional variable to lose/keep track of, just do all tests in the function in question

PR #56, opened 2019-06-26
circuit python support for azure iot hub commands

pelikhan:
-- Work in progress --
- [ ] send post message and check result


pelikhan:
going another route.

PR #57, opened 2019-07-07
Add Access point mode

mscosti:
- Adds access point creation methods to esp32spi.py.
  - Puts the ESP32 into AP mode so devices can connect to its own wifi local network.
  - Can specify the SSID and an optional password for the network

### Todo
- [x] More cleanup
- [x] Fix linting errors + get green build


ladyada:
i think this is a solid start. for making little servers, a 'micro Flask' could be a good way to create the structure for responding to queries. maybe we want to split this into two PRs, one for AP and one for server?

mscosti:
I think having 2 PRs is fine, one with AP, one with server. The only hesitation I have is that one isn't very useful w/out the other. Does this repo get a new release with every commit/PR, or is release cut manually when we think think its time?

In that case, I'll make this branch and PR just for AP and take out the example code. I'll open a new branch and draft PR for server with what I have now and can continue working on fleshing that out more.

ladyada:
we can make releases whenever. server is useful without AP (but agree AP is not useful without server) so lets do server first :)

mscosti:
Good to know. Server PR still has a lot more work to go into it, but I think this PR should be ready to go whenever we want, just because it's a simpler and more understood. 

We can sit on it until Server PR is merged.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -443,6 +468,15 @@ def is_connected(self):
             self.reset()
             return False
 
+    @property
+    def ap_listening(self):
+        """Whether the ESP32 is in access point mode and is listening for connections"""
```
brentru: "Returns if the ESP32 is in access point mode and is listening for connections"

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -473,6 +507,25 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
             raise RuntimeError("No such ssid", ssid)
         raise RuntimeError("Unknown error 0x%02X" % stat)
 
+    def create_ap(self, ssid, password, channel=b'\x01'):
```
brentru: Why do you need this method if you already have `wifi_set_ap_passphrase` and `wifi_set_ap_passphrase`? 

If those two methods are not meant to be user-accessable, and only called from within`create_ap`, consider changing them to private methods.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -473,6 +507,25 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+        """Create an access point with the given name and password."""
```
brentru: Document the kwargs for `create_ap` here.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -473,6 +507,25 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+        if isinstance(ssid, str):
```
brentru: Perform the this check within `set_ap_passphrase` instead, along with checking the length of the ssid.

adafruit_esp32spi/adafruit_esp32spi.py
brentru: I don't really like `channel` as a `byte`. Consider changing it to an `int` and then converting it into a byte within `create_ap`, along w/checking if it's btween 1 and 14.

mscosti: Yeah, I can change them to be private methods, they're not meant to be user called. This method (`create_ap`) will call `wifi_set_ap_passphrase` or `wifi_set_ap_network` depending on if you pass in a password or not to `create_ap`. The benefit here is after calling it will wait until the we see that esp32 has switches its status to `AP_LISTENING` to verify the call worked.

adafruit_esp32spi/adafruit_esp32spi.py
mscosti: will document all of the args 👍

adafruit_esp32spi/adafruit_esp32spi.py
brentru: Ok, I see the utility of the two sep. methods.

mscosti:
Thanks for the feedback @brentru ! I'll try to address your suggestions tomorrow night, unfortunately the only time I have to work on project is in my free time (nights, weekends) and this week I have a bit going on on.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -473,6 +507,25 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+            ssid = bytes(ssid, 'utf-8')
+        if password:
+            if isinstance(password, str):
+                password = bytes(password, 'utf-8')
+            self.wifi_set_ap_passphrase(ssid, password, channel)
+        else:
+            self.wifi_set_ap_network(ssid, channel)
+        for _ in range(10): # retries
```
brentru: You might want to implement your retry loop as time-based rather than retry-based. You may also want to consider making the number of retries (10) a variable kwarg named `retries`.

brentru:
@mscosti  No rush - great work so far!

adafruit_esp32spi/adafruit_esp32spi.py
mscosti: since `_wifi_set_ap_passphrase` and `_wifi_set_ap_network` are now both listed as private, I think it makes more sense to do all validation (including ssid/passphrase length checking and channel being 1-14) and arg documentation in `create_AP`, since that's the public method to be used.

This is also similar to the way `connect_AP` is written to provide a main entry point for calling either `wifi_set_passphrase` and `wifi_set_network`

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -462,17 +491,57 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
             self.wifi_set_passphrase(ssid, password)
         else:
             self.wifi_set_network(ssid)
-        for _ in range(10): # retries
+        times = time.monotonic()
+        while (time.monotonic() - times) < timeout_s:  # wait up until timeout
```
mscosti: switched `connect_AP` to also be timeout based instead of retry count based. Could speed up perceived connection time since we're checking more frequently now. (every 0.05 seconds as opposed to every 1 second)

adafruit_esp32spi/adafruit_esp32spi.py
mscosti: Great suggestion, thanks. Done.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -409,6 +411,29 @@ def wifi_set_entenable(self):
         if resp[0][0] != 1:
             raise RuntimeError("Failed to enable enterprise mode")
 
+    def wifi_set_ap_network(self, ssid, channel):
```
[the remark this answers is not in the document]
mscosti: decided to leave this the same to match the name of the command we send to the ESP32 FW.

mscosti:
@brentru Thanks for the review! I have updated to incorporate your feedback, and responded to some of them.

adafruit_esp32spi/adafruit_esp32spi.py
siddacious: This seems fine, may be useful as you say. How did you come to 0.05 seconds as a sleep time?

adafruit_esp32spi/adafruit_esp32spi.py
siddacious: I agree this makes more sense here. This is the function that the user interacts with and it makes sense that it holds the business  rules about error messages, etc.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -36,15 +36,16 @@
...
 class ESPSPI_WiFiManager:
     """
     A class to help manage the Wifi connection
     """
     NORMAL = const(1)
     ENTERPRISE = const(2)
 
-# pylint: disable=too-many-arguments
-    def __init__(self, esp, secrets, status_pixel=None, attempts=2, connection_type=NORMAL):
+    # pylint: disable=too-many-arguments, line-too-long
```
siddacious: No need to disable this lint check, just continue the arg list on the next line

adafruit_esp32spi/adafruit_esp32spi.py
mscosti: I came up with the value to match this: https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi.py#L175-L180

It seemed like a reasonable sleep time. I saw in other places 0.01 is used, but went with the slightly longer one.

I don't really have a preference for what time we use.

siddacious: This is fine

mscosti:
I think this PR also has all currently requested changes made, and is ready for potential re-review.

brentru:
@mscosti  Just tested this, I'd like to suggest some things before we merge:
- WiFiManager: `create_ap` should create an access point based on kwargs provided to it, a SSID and Password. I think people will use it this way
  - Providing the AP setup information to `create_ap` instead of setting it in `init`.
   - i.e: `wifi.create_ap(SSID, PASS)`
- Let's add an example of creating an access point so people don't need to dig through WiFiManager (Add two calls in the example and comment one out. Show creating an insecure access point and one for creating a password-protected access point)

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -127,6 +129,33 @@ def connect_normal(self):
                     self.reset()
                 continue
 
+    def create_ap(self):
```
brentru: I like this better: `create_AP(self, ssid, password)`. Dont need channel/timeout since they're already set in the esp32spi class.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -462,17 +491,57 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+        Create an access point with the given name, password, and channel.
+        Will wait until specified timeout seconds and return on success
+        or raise an exception on failure.
+
+        :param ssid: the SSID of the created Access Point. Must be 8 or more characters
+        :param passphrase: the password of the created Access Point. Must be 8 or more characters.
+        :param channel: channel of created Access Point (1 - 14).
+        :param timeout_s: number of seconds until we time out and fail to create AP
+        """
+        if len(ssid) < 8:
```
brentru: > A valid SSID is 0-32 octets with arbitrary contents. A 0-length SSID indicates the wildcard SSID (in probe request frames for instance).
You may want to limit the upper-bound of the SSID length, too

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -462,17 +491,57 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+            raise RuntimeError("ssid must be 8 more more characters")
+        if len(password) < 8:
```
brentru: > A pass-phrase is a sequence of between 8 and 63 ASCII-encoded characters. The limit of 63 comes from the desire to distinguish between a pass-phrase and a PSK displayed as 64 hexadecimal characters.

check that the password is password be ascii-encoded and doesnt go beyond 64 chars.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -462,17 +491,57 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+            raise RuntimeError("password must be 8 or more characters")
+        if channel < 1 or channel > 14:
+            raise RuntimeError("channel must be between 1 and 14")
+
+        if isinstance(channel, int):
```
brentru: Check `isinstance` attributes before the checks above?

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -462,17 +491,57 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+    def create_AP(self, ssid, password, channel=1, timeout_s=10): # pylint: disable=invalid-name
+        """
+        Create an access point with the given name, password, and channel.
+        Will wait until specified timeout seconds and return on success
+        or raise an exception on failure.
+
+        :param ssid: the SSID of the created Access Point. Must be 8 or more characters
+        :param passphrase: the password of the created Access Point. Must be 8 or more characters.
+        :param channel: channel of created Access Point (1 - 14).
+        :param timeout_s: number of seconds until we time out and fail to create AP
```
brentru: change timeout_s to timeout

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -462,17 +491,57 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
             raise RuntimeError("No such ssid", ssid)
         raise RuntimeError("Unknown error 0x%02X" % stat)
 
+    def create_AP(self, ssid, password, channel=1, timeout_s=10): # pylint: disable=invalid-name
+        """
+        Create an access point with the given name, password, and channel.
+        Will wait until specified timeout seconds and return on success
+        or raise an exception on failure.
+
+        :param ssid: the SSID of the created Access Point. Must be 8 or more characters
```
brentru: :param str ssid:

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -462,17 +491,57 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+        :param passphrase: the password of the created Access Point. Must be 8 or more characters.
```
brentru: :param str passphrase:

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -462,17 +491,57 @@ def connect_AP(self, ssid, password): # pylint: disable=invalid-name
...
+        :param channel: channel of created Access Point (1 - 14).
```
brentru: :param int channel:

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
mscosti: The reason I made it like this was to match the current pattern of the WifiManager, which requires a positional arg `secrets` object that contains the SSID and (now an optional) password. 

I figure since you can't currently be in both AP mode and Client mode at the same time, that it would make sense to re purpose the secrets object instance variable as a place to configure your AP SSID and optional password. 

Another reason (though I don't think this is a big deal in either case), is our example files wouldn't have to provide a default dummy SSID and password.

I don't disagree though, I just think it makes sense to be consistent.

mscosti:
I fixed a bug, and added AP creation examples to the wsgi server demo code as options to use instead of connecting to wifi.

I also modified the web application to no longer depend on externally hosted static assets, since if you're connected to the esp32 hotspot, you likely don't have a way for your device to download those assets from the internet unless your device has already cached them. Now it will work regardless.

@brentru I haven't updated the method signature for `create_AP` to take in kwargs for ssid and pass, and still use secrets for getting it.

The reason I think its better to use secrets is that the wifimanager class currently expects a secrets object to be passed to it, and every method in it uses that. If we took in kwargs, then there was no point in passing in secrets to the constructor. I thought about supporting some sort of dual mode where you could use kwargs *or* secrets, but it felt clunky to me. 

Curious what your thoughts are.

brentru:
@mscosti Had the chance to test this AM. Works great with each of the AP examples you provided.

> @brentru I haven't updated the method signature for create_AP to take in kwargs for ssid and pass, and still use secrets for getting it.

I like how AP creation works in your example - especially since it's contained within the init. for `wifimanager` as it's near the top of code near hardware initialization.

> I thought about supporting some sort of dual mode where you could use kwargs or secrets, but it felt clunky to me.

Agreed, the way you have it right now isn't clunky and we can always add a dual-mode in later. 

I'm going to merge this in and release, thank you for your patience with this PR.

PR #58, opened 2019-07-11
Add Server functionality

mscosti:
This is still a work in progress, but opening it up for early feedback and suggestions.

- Adds Server creation methods to esp32spi.py
  - Instantiates a server on the esp32 on the specified port and a specified socket that was previously requested
- Adds a new example file that creates an access point from secrets.py and then a server on port 80
  - Logs when a device connects or disconnects from the access point
  - When a device hits `<serverIP>:80` (like in their browser), send back simple HTML page that has a couple of anchor tags on it

### Todo
- [ ] rebase with master once #57 is merged
- [ ] Make the example code more polished and more fun. 
- [ ] Create a new helper class to manage server creation and management. 
- [ ] Thinking we could even go so far as to implement simple request handling. Something like:
   ```
      Server.on("<HttpMethod>", "</routeString>", callbackFn)
      ... 
      def callbackFn(data, socketToWriteTo):
   ```
   - This Might be confusing since the run loop would need to be handled outside code.py?
   - At the very least could have a method to check *if* a certain route is triggered?
   - Could instead have a more feature rich web server lib (like flask) live outside of this to keep it focused.
- [ ] More cleanup
- [ ] More testing
- [ ] Fix linting errors + get green build

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -66,6 +66,7 @@ _GET_CURR_RSSI_CMD     = const(0x25)
 _GET_CURR_ENCT_CMD     = const(0x26)
 
 _SCAN_NETWORKS         = const(0x27)
+_START_SERVER_TCP_CMD  = const(0x28)
 _GET_SOCKET_CMD        = const(0x3F)
 _GET_STATE_TCP_CMD     = const(0x29)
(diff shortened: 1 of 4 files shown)


siddacious:
@mscosti Are you going to open a new PR for this?

siddacious:
Yes you are/did! Never mind me :)

PR #59, opened 2019-07-11
Add Server creation and management support

mscosti:
This is still a work in progress, but opening it up for early feedback and suggestions.

## Core Library additions
- Adds Server creation methods to esp32spi.py
  - Instantiates a server on the esp32 on the specified port and a specified socket that was previously requested
- Adds `WSGIServer` class
  - This is a simple [WSGI](https://www.python.org/dev/peps/pep-0333/) compatible server implemtnation
  - Accepts an application callable to be passed to it.
  - Invokes the application callable when a request comes in
      - Passes an `environment` dictionary with request values per wsgi requirements.
      - Passes a `start_response` method that must be invoked before the `application` callable returns
      - Application callable must return with an iterable form of response data. 

## New example file
- Shows how to write a simple WSGI compatible application callable, in the form of a callable class
   - supports basic request handling
   - supports serving files from disk
- Shows how to use the example Application class to write a simple web app.
   - serves an index.html and JS file from disk when hitting the server from your browser
   - adds an `led_on` and `led_off` handler to turn on /off an LED
   - adds in `ajax/led_color` handler to set the LED color

## Todo outside of this PR
Start discussing what a Circuit Python WSGI web application framework might look like.
  - Would live in a separate repository, and would be compatible with any WSGI Server


siddacious:
Looks like you ran into some pylint issues. Let me know if you need some assistance in resolving them

mscosti:
Thanks. I installed an in editor extension to lint as a type so shouldn't be seeing anymore of those errors moving forward.

This PR is going to change a lot, as I originally had Server + AP in one, but we're going to be splitting them up. Since Server code is going to go in first, I'll need to do some git manipulation to get this PR to be independent of any AP code (#57)

ladyada:
thank you for the effort, we'll get everything going, together!
let us know if you need a hand. you may want to start 2 new branches?

arturo182:
This is great, something I wanted to look into as well, thanks for doing this!

A request handling like you suggested would be great, I think it should also be possible to serve .html files directly from the filesystem.

Also, having a simple template system, but that's a separate library, I think.

mscosti:
Yeah, definitely want to look into serving .html files from filesystem so you don't need to inline it in your python code, but I'm not sure yet if this would be a something built in to the lib or not. I also agree that a templating system would be separate.

Next up I think is to figure out what the request handling looks like. I'm thinking `Server` will maybe support two "modes". 

1) In one mode you can register all your request routes and request handlers upfront before starting the server, and they will be automatically called when that particular request comes in. Unless there is something I'm not thinking of right now, I think this would only be possible if the server is in control of the main event loop. So it's simple to use for responding to requests, but you would only be act on that.

2) The other mode I'm thinking would be more manual and more user code, but with it more flexibility. You would have control over the main event loop in your code, but you'd be responsible for asking `Server` class to see if an incoming request matches a specific route, then handle it. This should make it so you can still do things outside of handling a specific event, like reading sensors or what have you.

I'm curious if its possible to support both modes at the same time if you have the server request listening loop and callback handling on a separate thread from code.py? I'm not familiar with threading in python, let alone know what is possible in circuitpython/micropython, though.

mscosti:
@arturo182  I agree with you that static asset handling is an important enough feature for getting a useful webserver running quickly. 

I've made it so you can configure a single directory as your 'static assets' folders by doing
```
server.set_static_dir("/path/to/folder")
```
and if a request comes through that doesn't match any of your own configured request handlers, but *does* match a GET request for one of the files in that directory, then that file will get automatically returned. 

It tries to determine the content-type based on the file extension. Right now I've only tested .html, .css, and .js, but images should work in theory, but might need some special handling.

I've also made the method `server.serve_file(self, file_path, dir=None)` public so you can still have writing files to the socket handled for you in your own request handlers

adafruit_esp32spi/adafruit_esp32spi_socket.py
```diff
@@ -148,6 +151,39 @@ def settimeout(self, value):
...
+
+    def connected(self):
+        """Whether or not we are connected to the socket"""
+        if self.socknum == NO_SOCKET_AVAIL:
+            return False
+        elif self.available():
+            return True
+        else:
+            status = _the_interface.socket_status(self.socknum)
+            # TODO: why is esp.<ConstantName> not defined? using magic numbers in mean time
```
mscosti: I wanted to use the same constants as defined in `adafruit_esp32spi`, but for some reason I wasn't able to get that to work. The constants weren't accessible off of the that module when running in CP (but my editor was showing them as available)

Any ideas? We obviously don't want the magic numbers here.

adafruit_esp32spi/adafruit_esp32spi_socket.py
```diff
@@ -148,6 +151,39 @@ def settimeout(self, value):
...
+            result = status not in (1,
```
mscosti: for context, stole this logic from the arduino `WiFiNINA` lib: https://github.com/arduino-libraries/WiFiNINA/blob/master/src/WiFiClient.cpp#L221-L242

examples/static/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.4.1/jquery.min.js"></script>
+        <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery-minicolors/2.3.4/jquery.minicolors.min.js"></script>
+        <script async src="led_color_picker_example.js"></script>
+        <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/jquery-minicolors/2.3.4/jquery.minicolors.css" />
+        <link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/4.3.1/css/bootstrap.min.css" />
```
mscosti: I'm not familiarity with how licensing works with open source projects using other libraries. I double checked that jQuery, jQuery minicolors plugin, and bootstrap are all on MIT license, at least. 

Let me know if this is not sufficient, we can re-write the the frontend to not need 3rd party libs.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -54,6 +54,7 @@
 # pylint: disable=bad-whitespace
 _SET_NET_CMD           = const(0x10)
 _SET_PASSPHRASE_CMD    = const(0x11)
+_SET_AP_PASSPHRASE_CMD = const(0x19)
```
mscosti: woops, this is copy paste from my AP pull request. This can be removed from this PR.

adafruit_esp32spi/adafruit_esp32spi_server.py
```diff
@@ -0,0 +1,193 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2019 ladyada for Adafruit Industries
```
sommersoft: Change the copyright to yourself. You deserve it! 😄

adafruit_esp32spi/adafruit_esp32spi_server.py
```diff
@@ -0,0 +1,193 @@
...
+import adafruit_esp32spi.adafruit_esp32spi_socket as socket
+from adafruit_esp32spi.adafruit_esp32spi_requests import parse_headers
+
+_the_interface = None   # pylint: disable=invalid-name
+def set_interface(iface):
+    """Helper to set the global internet interface"""
+    global _the_interface   # pylint: disable=global-statement, invalid-name
+    _the_interface = iface
+    socket.set_interface(iface)
+
```
sommersoft: The use of a global variable always raises a flag for me, personally. However, I'm currently at a loss of how to address this situation differently. I agree that it shouldn't be instanced in the `server` class...

Anyone in @adafruit/circuitpythonlibrarians have an idea to avoid this global?

adafruit_esp32spi/adafruit_esp32spi_server.py
```diff
@@ -0,0 +1,193 @@
...
+NO_SOCK_AVAIL = const(255)
+INDEX_HTML = "/index.html"
+
+
+# pylint: disable=unused-argument, redefined-builtin, invalid-name
+class server:
```
sommersoft: Class name should be `Server` to meet PEP8.

adafruit_esp32spi/adafruit_esp32spi_server.py
```diff
@@ -0,0 +1,193 @@
...
+        request_handler should accept the following args:
+            (Dict headers, bytes body, Socket client)
+
+        :param str method: the method of the HTTP request
+        :param str path: the path of the HTTP request
+        :param func request_handler: the function to call
+        """
+        self._listeners[self._get_listener_key(method, path)] = request_handler
+
+    def set_static_dir(self, directory_path):
```
sommersoft: Again, we try to avoid `get` and `set` method names. This could be a property:
```python
@property
def static_dir(self):
    return self._static_dir

@static_dir.setter
def static_dir(self, directory_path):
   # current 'set_static_dir'
```

siddacious:
Thanks @mscosti ! This is all great work, definitely will be very useful and make it much easier to setup REST-ey or webservices or pages to interact with the hardware.

My main concern right now is if the SPI32SPI library is the place to have a webserver live. It seems like  this library should handle setting up the network connection and listening on a given port, but a separate library should be responsible for accepting and processing requests.

Great work, I'm really looking forward to using this and helping out round out the feature set, but I think we should make sure it's being put in the right place.

mscosti:
I don't  disagree @siddacious , but I think it's a larger discussion. You're right that this adds a lot of functionality that is definitely not a part of purely sending and receiving commands to/from the esp32 over SPI. We have the the 'adafruit_esp32spi.py' class for this.

But, this library has already gone past that line by introducing helper classes for making HTTP requests easier and parsing the response out into a response class (headers and body separated and all).

So, I do agree with you, but if that's the line we want to draw than I think we would then also have to move over everything else that has an understanding of the HTTP protocol (incoming HTTP request handling/routing and responding. Outgoing request creation and response parsing). 

I can also see that server could be a little different because the feature set _could_  grow pretty large, but it also doesn't have to. We just need to agree on where that line is drawn.  Happy to continue to the discussion, I want this to end up in the right place too 🙂

siddacious:
You're right @mscosti this certainly is a part of a larger discussion. This lib has definitely grown past purely talking to the ESP32 and has started including useful helpers for managing a wifi connection, making HTTP requests and packaging the responses into response objects as you say. While this is indeed beyond the obvious scope of "manage talking to the ESP32", what's been added is useful and for now at least makes sense to live here. 

I think the dividing line that makes sense to me is to have this library do an incoming/server side equivalent to the outgoing/client side helpers that already exist. The way I see it, the existing helpers allow the user to create an HTTP request , then it will send the request and package up the response for the user.

In my mind, the correlate to that would be allowing the user to create an endpoint for clients to call, package up incoming requests into an object that something can be done with, and then handle returning the response to the user.

In terms of how I would see that changing what you've already done, I would imagine that the `server` object that this library creates would have an attribute like `request_processor` or similar that takes a single handler function with a defined interface. The handler function would be passed a request object and would return a response object. Calling `update_poll` in the main loop would call the handler in response to new requests. Basically as it is now but there would only be one `on` endpoint. Another library could handle farming that request out to different handlers depending on method, URI, etc.

adafruit_esp32spi/adafruit_esp32spi_server.py
arturo182: This is same as in the other files in this library (see https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi_socket.py#L37), so I think it's ok for now, we might fix it later in all the classes.

arturo182:
I should have some time tomorrow to verify this on HW :)

mscosti:
Thanks @arturo182 ! :smile:

adafruit_esp32spi/adafruit_esp32spi_server.py
sommersoft: Fair point @arturo182. I obviously didn't look to see that it was already in the lib.

arturo182:
Oh damn, it slipped my mind! Will do it tomorrow for sure 😅

mscosti:
No rush @arturo182,  my weeknights this week are a bit busy so I won't be able to do much until tomorrow and Friday, and I have to address some PR comments and possibly some re working depending on what feature set we want in this library anyway.

Speaking of which:

@siddacious I think I'm understanding you correctly but want to make sure and to clarify. I think what you're suggesting is that `Server` would accept a generic `request_handler` function. This would be called from a stripped down version of `update_poll` whenever a request comes in, where it would simply pass on some parsed data, maybe something like a `Request` object that has
1. A parsed header dict
2. The raw body content
3. The request method (`GET`, `POST`, etc)
4. The request path
It wouldn't act on any of the above, just pass it along to the registered `request_handler`

That would then mean that the "endpoint registering and management" (on `GET /my-endpoint` call `getFunc`,  on `POST /my-endpoint` call `postFunc`), could be moved somewhere else, along with auto static asset serving among other things and future features. It would be a pretty smooth transition to move the extra functionality that I already have to wherever we want to move it.

I guess, my only concern/question is, where do we think is the best place to put in all the extra server management goodies (both what we have right now + future features)? Would it be a new adafruit CP lib? The esp32SPI `Server.request_handler` would allow for pretty basic and unbiased flexible use, but I would imagine that most users will want this secondary library in order to get the more 'complete' server experience. 

If we spin up a new lib, I feel like we would need a lot more design discussions to make sure we get it right. Another thing to think about is while it would be easy to just slot in what we have as middleware that goes on top of ESP32SPI (so it would likely have a dependency on ESP32SPI), if we put the effort in to make the new library I feel like it it *shouldn't* actually have the dependency at all and would need to be agnostic; the hardware api would need to conform to a strict interface. Otherwise there would be little gain in separating them out in the first place. Are there more circuit python wifi chips/coprocessors that can act as a server and would thus benefit from a generic server library?

Sorry for the long train of thought post and if i'm over analyzing our options 😅 I'm just really torn between what is maybe more correct from an engineering perspective and with what provides the best user experience for the ESP32 airlift boards. We both already acknowledged it's a big discussion, so just adding more questions and things to think about to it. I'm curious what a path forward might look like.

siddacious:
@mscosti it sounds like you understood my suggestion perfectly, based on your summary. In all candor I'm leaning on my experience with the Rack framework in Ruby:
https://www.engineyard.com/blog/understanding-rack-apps-and-middleware

Rack defines a simple but robust interface for web apps or middleware to plug into a request handling stack on top of a range of different app servers. The whole middleware stack is way beyond what I have in mind which is more the clearly defined interface as you suggested.

I just did a bit of looking and of course there is an equivalent standard in Python, WSGI:
http://ivory.idyll.org/articles/wsgi-intro/what-is-wsgi.html
We can probably take a lot of inspiration from it, if not steal/use the idea outright. 

With regards to where the extra server stuff would go, I was imagining a separate lib. I agree that if the intent is for it to be transport/hardware agnostic it would make sense to have a strictly defined interface. Perhaps I'm being naive but I don't think we need to rack our brains too hard over the API. WSGI, Rack or something similar would be a perfectly good first pass and we could refine it later.

The only other socket providing thing I know is somewhere near the horizon is the Ethernet FeatherWing though I imagine others are possible. It's certainly possible, even likely that Airlift/ESP32SPI would be the only hardware supporting the server library for a while. With that in mind I wouldn't be entirely opposed to keeping them together for the time being, though they would still want to be internally fairly distinct. 

Thanks for all the thought (and work!) you're putting into this. If you're feeling torn between between doing things "correctly" and what provides the best UI, in my opinion UI wins every time. I would suggest trying to avoid getting too caught up trying to find an optimal solution; better IMHO to iterate more quickly, fail early, and roll with the punches as they come.

I'm happy to opine more but unless you have more questions for me, I'm content to leave it up to you to figure out the best approach. Other's opinions are always welcome and valued.

arturo182:
I'm going away for vacation for 2 weeks tomorrow so really wanted to test this today. 

Works for me! 

https://twitter.com/arturo182/status/1151491502478168064?s=19

examples/esp32spi_server.py
```diff
@@ -0,0 +1,83 @@
...
+server.set_interface(esp)
+server = server.server(80, debug=False)
+
+
+def onLedHigh(headers, body, client): # pylint: disable=unused-argument
+    print("led on!")
+    status_light.fill((0, 0, 100))
+    server.serve_file("static/index.html")
+
+def onLedLow(headers, body, client): # pylint: disable=unused-argument
```
brentru: Change to `led_off`

mscosti:
@siddacious Wow, thank's so much for bringing these concepts and standard interfaces to my attention. *Ofcourse* there are standard interfaces for application frameworks to talk with app servers. It seems way too obvious now 😅 

To me, the WSGI standard seems like it is ***exactly*** what we want this to look like. Any future boards that provide sockets would just need to write a server with the WSGI api, and we could create a general Circuit Python WSGI web application framework that should just work with it right away. I'm on board, and will start working on making the server WSGI compliant 😄 

For the sake of quick iteration like you said, providing a good experience, and illustration of how it all ties together, I think it will also be helpful to include a simple WSGI framework in this lib to use with the WSGI server until we actually have a proper Circuit Python implementation ready.

examples/esp32spi_server.py
```diff
@@ -0,0 +1,83 @@
...
+
+## Connect to wifi with secrets
+wifi = wifimanager.ESPSPI_WiFiManager(esp, secrets, status_light, debug=True)
+wifi.connect()
+
+server.set_interface(esp)
+server = server.server(80, debug=False)
+
+
+def onLedHigh(headers, body, client): # pylint: disable=unused-argument
```
brentru: Lets make this more pythonic: `led_on`

examples/esp32spi_server.py
```diff
@@ -0,0 +1,83 @@
...
+
+def onLedLow(headers, body, client): # pylint: disable=unused-argument
+    print("led off!")
+    status_light.fill(0)
+    server.serve_file("static/index.html")
+
+def onLedColor(headers, body, client): # pylint: disable=unused-argument
+    rgb = json_module.loads(body)
+    print("led color: " + rgb)
+    status_light.fill((rgb.get("r"), rgb.get("g"), rgb.get("b")))
```
brentru: You might want to perform the `.fill(r,g,b)` a bit cleaner

examples/esp32spi_server.py
```diff
@@ -0,0 +1,83 @@
...
+    client.write(b"HTTP/1.1 200 OK")
+    client.close()
+
+server.set_static_dir("/static")
+server.on("GET", "/H", onLedHigh)
```
brentru: Possibly we'd attach the methods like how method callbacks are handled in MiniMQTT.

https://github.com/brentru/Adafruit_CircuitPython_MiniMQTT/blob/master/examples/minimqtt_simpletest.py#L93

siddacious:
@mscosti Awesome, sounds great. For the WSGI framework in this lib to test with, I would have it as an example rather than having it live in the main codebase.

examples/esp32spi_server.py
```diff
@@ -0,0 +1,83 @@
...
+server.on("GET", "/L", onLedLow)
+server.on("POST", "/ajax/ledcolor", onLedColor)
+
+
+print("open this IP in your browser: ", esp.pretty_ip(esp.ip_address))
+
+server.start()
+
+while True:
+    server.update_poll()
```
brentru: update_poll is dependent on if ESP32SPI's connection doesn't drop out and toss a `RuntimeError`.

You'll want to reset the `wifi` manager if this occurs within your loop.

https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/examples/esp32spi_aio_post.py#L59

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -49,13 +50,14 @@ def __init__(self, esp, secrets, status_pixel=None, attempts=2):
...
         self.esp = esp
-        self.debug = False
+        self.debug = debug
         self.ssid = secrets['ssid']
         self.password = secrets['password']
         self.attempts = attempts
         requests.set_interface(self.esp)
         self.statuspix = status_pixel
         self.pixel_status(0)
+    # pylint: enable=too-many-arguments
```
brentru: extra pylint directive here?

mscosti:
That's a fair compromise. Sounds good. 👍

adafruit_esp32spi/adafruit_esp32spi_socket.py
brentru: Which line are you referring to here?

mscosti: all of the numbers in `status not in (magic numbers)`

All of those are constants corresponding to ones defined in esp32spi.py https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi.py#L101-L111

My initial attempt to use them wasn't working, I'm sure there is just something silly i'm missing with being able to re-use those constants here.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
mscosti: I don't think so, unless i'm misunderstanding the question. I disabled `too-many`arguments on line 41 because I updated __init__ to accept the debug boolean, pushing it over the pylint suggested arg limit, so I'm re-enabling that linting rule after the __init__ function so it applies to every other method in this file.

siddacious: This is fine, you're understanding correctly.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -622,6 +624,32 @@ def socket_close(self, socket_num):
...
+                                                    (conn_mode,)))
+        else:       # use the 3 arg version
+            resp = self._send_command_get_response(_START_SERVER_TCP_CMD,
+                                                   (port_param,
+                                                    self._socknum_ll[0],
+                                                    (conn_mode,)))
+        if resp[0][0] != 1:
+            raise RuntimeError("Could not start server")
+
+    def get_server_state(self, socket_num):
```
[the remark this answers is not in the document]
mscosti: I think in this situation that wouldn't work, because it takes in a parameter? I will still change it to server_state though, to better match similar status queries (like socket_available, which takes the socknum and returns available bytes)

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -622,6 +624,32 @@ def socket_close(self, socket_num):
...
+        """Get the state of the ESP32's internal reference server socket number"""
+        self._socknum_ll[0][0] = socket_num
```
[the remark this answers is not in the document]
mscosti: To be completely honest, I'm not 100% sure what the instance variable is used for, but it seems to be a pattern that whenever an internal socket num reference is passed in, we update `self._socknum_ll[0][0]` to it.

I continued to follow the pattern for consistency, but i'm also curious about its purpose. It seems everywhere its used we are writing to it and then reading from it in the same method; never only reading from it.

examples/esp32spi_server.py
mscosti: *Note*: moved to `examples/esp32_wsgiserver.py`

If I'm understanding your example, there are a discrete set of event handlers that the MiniMQTT client expects, so you can just assign callbacks to those specific properties and the client knows to call those under specific circumstances.

In the web application scenario, the circumstances for when a callback is called is not fixed and is defined by the user creating the web application. ex, I want this callback (`led_on`) to be called when a `GET` request comes in to the path `/H`. 

So the event registration needs to somehow accept these 3 arguments. If there is a better way to do this, please let me know. 

I don't think it's as crucial though, because now this no longer a part of the library itself, and is only used for illustration purposes in the example files. Someone could create their own wsgi application to use with the lib provided  wsgi server that behaves and looks entirely different.

examples/esp32spi_server.py
mscosti: resolved in examples/esp32_wsgiserver.py

examples/esp32spi_server.py
mscosti: resolved in `examples/esp32_wsgiserver.py`

examples/esp32spi_server.py
mscosti: resolved in `examples/esp32_wsgiserver.py`

examples/esp32spi_server.py
mscosti: resolved in `examples/esp32_wsgiserver.py`

mscosti:
@siddacious @brentru @sommersoft , I have completed the refactoring effort to have the main functionality that this lib provides to be a WSGI compatible server implementation, as discussed. 

https://www.python.org/dev/peps/pep-0333/

The examples folder now contains an example that includes a simple 'web application framework' to illustrate how to interface with the wsgi server. It shows how to write an application with basic custom request handlers and file serving.

I have also addressed all PR comments with updated code or a follow up question/answer.

adafruit_esp32spi/adafruit_esp32spi.py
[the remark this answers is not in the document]
sommersoft: 🤦‍♂ Not my most complete thought, obviously. haha

examples/esp32spi_server.py
siddacious: This is good; users can register handlers however they wish in their app/server that consumes the WSGI interface (interface interface, ya I know)

examples/esp32spi_wsgiserver.py
```diff
@@ -0,0 +1,188 @@
...
+import board
+import busio
+from digitalio import DigitalInOut
+import neopixel
+
+from adafruit_esp32spi import adafruit_esp32spi
+import adafruit_esp32spi.adafruit_esp32spi_wifimanager as wifimanager
+import adafruit_esp32spi.adafruit_esp32spi_wsgiserver as server
+
+# This example depends on the 'static' folder in the examples folder
```
brentru: Check for the existence of a static folder, or at least a file within `static/*.html`

See:
https://github.com/adafruit/Adafruit_CircuitPython_Display_Button/blob/master/examples/display_button_simpletest.py#L17

examples/esp32spi_wsgiserver.py
```diff
@@ -0,0 +1,188 @@
...
+    raise
+
+try:
+    import json as json_module
+except ImportError:
+    import ujson as json_module
+
+print("ESP32 SPI simple web server test!")
+
+esp32_cs = DigitalInOut(board.D10)
```
brentru: Add pin defs for boards with builtin ESP32 like the PyPortal, retain the pin defs you have for external ESP32 breakouts

Example: https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/examples/esp32spi_aio_post.py#L18

examples/esp32spi_wsgiserver.py
```diff
@@ -0,0 +1,188 @@
...
+wifi = wifimanager.ESPSPI_WiFiManager(esp, secrets, status_light)
+wifi.connect()
+
+class SimpleWSGIApplication:
+    """
+    An example of a simple WSGI Application that supports
+    basic route handling and static asset file serving for common file types
+    """
+
+    INDEX = "/index.html"
```
brentru: If you're requiring an index.html, it should be checked. See above comment on how to do this. Sometimes files dont get copied over completely to the filesystem and people don't notice it.

examples/esp32spi_wsgiserver.py
```diff
@@ -0,0 +1,188 @@
...
+status_light = neopixel.NeoPixel(board.NEOPIXEL, 1, brightness=0.2) # Uncomment for Most Boards
+"""Uncomment below for ItsyBitsy M4"""
+# import adafruit_dotstar as dotstar
+# status_light = dotstar.DotStar(board.APA102_SCK, board.APA102_MOSI, 1, brightness=1)
+
+## Connect to wifi with secrets
+wifi = wifimanager.ESPSPI_WiFiManager(esp, secrets, status_light)
+wifi.connect()
+
+class SimpleWSGIApplication:
```
brentru: Would it be better to make a library to handle this class, called `CircuitPython_WSGI`?

examples/esp32spi_wsgiserver.py
```diff
@@ -0,0 +1,188 @@
...
+            (status, header_list, data_iterable)
+
+        :param str method: the method of the HTTP request
+        :param str path: the path of the HTTP request
+        :param func request_handler: the function to call
+        """
+        self._listeners[self._get_listener_key(method, path)] = request_handler
+
+    def serve_file(self, file_path, directory=None):
+        status = "200 OK"
```
brentru: If you're going to move this class into a module, something to consider would be making an dict holding all the HTTP status codes.

examples/esp32spi_wsgiserver.py
mscosti: yep, absolutely. With a new repo/library there are probably a lot more features we could add to make building web apps really simple. This is something I'd like to contribute to. 

This is just a simple example of how to use the wsgi server that the esp32SPI library provides. It's possible we remove this part of the example from here when we have a separate and more full featured wsgi web app library. Until we have our own solution built though, I think its important to at least illustrate how to write a wsgi app; show that its really not that complicated, here is something to use as is for simple use case, or here is a starter base to add your own features on top of. 

To me, that's the best part about using an interface like wsgi to define communication between application and server.

examples/esp32spi_wsgiserver.py
mscosti: Good point, something like this would be a nice convenience to have in a more full featured library.

For this simple example, I think we're fine to leave that feature out.

examples/esp32spi_wsgiserver.py
ladyada: you got it!
https://github.com/adafruit/Adafruit_CircuitPython_WSGI

examples/esp32spi_wsgiserver.py
mscosti: Checking now before starting the server.

mscosti:
@brentru I have updated the PR with your feedback and is ready for re review whenever you have the chance

brentru:
@mscosti  Thank you for making the requested changes - the example looks good to me.

mscosti:
Thanks @brentru !

Does anyone else have any additional feedback to add? @siddacious @sommersoft @ladyada ?
Do we need another round testing since a fair amount has changed since @arturo182 tested it? 

I'm also curious what the process is for getting things squashed and merged, and then released?

brentru:
@mscosti  I'll do a test on a PyPortal later today with the current PR commit.

brentru:
@mscosti  Could you put the esp32spi_wsgiserver.py and /static/ into one folder in `examples/server`?

Works perfectly on a PyPortal running CircuitPython 4.0.2.

mscosti:
@brentru Thanks for testing! 

Moved static folder and example py script to a new folder located at `/examples/server`

brentru:
@mscosti Looks OK to me for a merge into `master` if others agree. I was able to get the example up and running quickly on a PyPortal.

After this PR merges in, let's chat on discord about building out the WSGI library next  (https://github.com/adafruit/Adafruit_CircuitPython_WSGI).

siddacious:
Looks great! Merging!

PR #65, opened 2019-08-21
Fixed MAC reverse byte order issue.

mytechnotalent:
Fixed MAC reverse byte order issue within the ESP32SPI library as it was originally reporting the MAC address in reverse byte order. Added portable MAC function call within WSGISERVER.PY example to properly echo MAC address when running example.


mytechnotalent:
Re-Submitting.

PR #66, opened 2019-08-21
Fixed MAC reverse byte order issue within the ESP32SPI library as it …

mytechnotalent:
…was originally echoing the MAC address in reverse byte order.  Added portable MAC function call within WSGISERVER.PY example to properly echo MAC address when running example.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -341,7 +341,7 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
 
     @property
     def MAC_address_actual(self):        # pylint: disable=invalid-name
-        """A bytearray containing the MAC address of the ESP32"""
+        """A bytearray containing the actual MAC address of the ESP32"""
         if self._debug:
             print("MAC address")
         resp = self._send_command_get_response(_GET_MACADDR_CMD, [b'\xFF'])
@@ -770,4 +770,3 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
                                                ((pin,), (value,)))
         if resp[0][0] != 1:
             raise RuntimeError("Failed to write to pin")
-            
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -337,6 +337,14 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
         if self._debug:
             print("MAC address")
         resp = self._send_command_get_response(_GET_MACADDR_CMD, [b'\xFF'])
+        return resp[0]
+
+    @property
+    def MAC_address_actual(self):        # pylint: disable=invalid-name
+        """A bytearray containing the MAC address of the ESP32"""
+        if self._debug:
+            print("MAC address")
+        resp = self._send_command_get_response(_GET_MACADDR_CMD, [b'\xFF'])
         new_resp = bytearray(resp[0])
         new_resp = reversed(new_resp)
         return new_resp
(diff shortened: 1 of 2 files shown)


mytechnotalent:
@ladyada updated code as requested however Travis seems to be failing on build.

mytechnotalent:
@ladyada ganbatte done it :)

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -337,7 +337,9 @@ def MAC_address(self):        # pylint: disable=invalid-name
         if self._debug:
             print("MAC address")
         resp = self._send_command_get_response(_GET_MACADDR_CMD, [b'\xFF'])
-        return resp[0]
+        new_resp = bytearray(resp[0])
```
ladyada: check it! you can turn these three lines into one line of code by nesting the commands. please try it!

mytechnotalent: @ladyada not sure what you are asking as the other functions follow the same format in line 337, 338.  Lines 337, 338, 339 are part of the original code so not sure where the efficiency is here.

ladyada: its not required, but sometimes people like to shorten it. how are you verifying the MAC address - with a router or...?

mytechnotalent: @ladyada understood but the rest of the functions follow that convention so wanted to stay consistent for style as that would really freak me out seeing every other function in one coding style and this one in another format UBER OCD HERE!  My home router I only allow devices to which I specially add their MAC.  When I reversed the byte order I was successfully able to connect.   To confirm I also fired up my Kali box and did a netdiscover and clearly verified the reverse order sequence and the nice Expressif Inc. Vendor tag.

ladyada: oki!

ladyada:
@brentru wanna try this and use an external tool to verify the MAC of the esp32

anecdata:
minor point, but can we mark this as a breaking change in the release notes? Some people (like me) may have code that uses the old order.

ladyada:
why would it be a breaking change? usually you display the MAC, but code doesnt stop working because of the new order

anecdata:
I suppose it's not breaking within CP (and perhaps dumb on my part), but for example, I use an ID based on old order to ID the ESP unit to my server.

ladyada:
whatever y'all want to do. they're just numbers

mytechnotalent:
I don't want to break anyone's code.  What I will do is add a new function to the **adafruit_esp32spi.py** library called **MAC_address_actual** to which you as the community can have a review on.

mscosti:
For what its worth, it appears that the nina-fw is purposely sending the order this way, and the arduino WiFiNina library doesn't do anything to adjust it. In the [example usage](https://www.arduino.cc/en/Reference/WiFiNINAMACAddress) you can see they're reversing the output of the call when printing it, but the byte array the `.macAddress()` function returns is left untouched. 

I'm not sure why this is the way it is. Does it make sense to simply document this behavior, and perhaps implement a helper method that reverses it for use when printing?

ladyada:
i think reversing it in the example is also fine.

kattni:
@brentru Can you test this if you did not already? It looks like it's otherwise ready to merge. Thanks!

brentru:
@kattni  I'll test later today. Assigning myself....

kattni:
@brentru I believe it's possible that since the changes don't conflict with that part of the file, merging this PR will not have any problems. If you're satisfied with the changes, I say we merge it and verify that the newer functionality is still in the file. We can always revert it if necessary.

brentru:
@kattni  Sounds good, I have no problems with the changes and would be OK to merge.

kattni:
@brentru Please merge it and verify the functions in question are still present. I will release the library if you give the go ahead.

FoamyGuy:
I verified there were no merging conflicts. Also tested the changes afterward. The new property and edited example are working as expected. As a sanity check I also tested the updated version with an example unrelated to this change and everything checked out fine there as well.

PR #67, opened 2019-08-22
Expose nina-fw's getTime

brentru:
Exposing [getTime function](https://github.com/adafruit/nina-fw/blob/master/main/CommandHandler.cpp#L792) to fetch latest unix timestamp.

`get_time` will raise an error if the ESP32 module is not connected to WiFi before the method is called.


mscosti:
I'm assuming this doesn't work if you're in AP mode and do not have an internet connection? or does the ESP32 actually have a hardware clock?

ladyada:
it does ntp so yes network required!

brentru:
@mscosti  I didn't implement a check for AP mode, only if it's connected to WiFi. Do you want to add a PR to raise an error if it's in AP mode?

The ESP32 doesn't have an internal clock, the NTP is set to a fixed url here: https://github.com/adafruit/nina-fw/blob/master/arduino/libraries/WiFi/src/WiFi.cpp#L577

mscosti:
:+1: Sure thing, I'll see what I can do.

PR #69, opened 2019-08-24
Adding to the functionality of the ESP32 library in order to provide …

mytechnotalent:
…a printable echo of a remote IP connection in addition to optional code to control specific IP access to the server as the server can only take one connection at a time.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -82,6 +82,7 @@ _REQ_HOST_BY_NAME_CMD  = const(0x34)
 _GET_HOST_BY_NAME_CMD  = const(0x35)
 _START_SCAN_NETWORKS   = const(0x36)
 _GET_FW_VERSION_CMD    = const(0x37)
+_GET_REMOTE_DATA_CMD   = const(0x3A)
 _GET_TIME              = const(0x3B)
 _PING_CMD              = const(0x3E)
 
@@ -768,3 +769,8 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
             resp = self._send_command_get_response(_GET_TIME)
             return struct.unpack('<i', resp[0])
         raise RuntimeError("Must be connected to WiFi before obtaining NTP.")
+
+    def get_remote_data(self, socket_num):
+        self._socknum_ll[0][0] = socket_num
+        resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD, self._socknum_ll, reply_params=2)
+        return resp[0]
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -770,8 +770,10 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
             return struct.unpack('<i', resp[0])
         raise RuntimeError("Must be connected to WiFi before obtaining NTP.")
 
+    """ Method to obtain remote IP """
     def get_remote_data(self, socket_num):
         self._socknum_ll[0][0] = socket_num
         resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD,
-                                self._socknum_ll, reply_params=2)
+            self._socknum_ll, reply_params=2)
         return resp[0]
+
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -770,10 +770,11 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
             return struct.unpack('<i', resp[0])
         raise RuntimeError("Must be connected to WiFi before obtaining NTP.")
 
-    """ Method to obtain remote IP """
     def get_remote_data(self, socket_num):
+        """
+        Method to obtain remote IP
+        """
         self._socknum_ll[0][0] = socket_num
         resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD,
-            self._socknum_ll, reply_params=2)
+                                                self._socknum_ll, reply_params=2)
         return resp[0]
-
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -776,5 +776,5 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
         """
         self._socknum_ll[0][0] = socket_num
         resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD,
-                                                self._socknum_ll, reply_params=2)
+                                               self._socknum_ll, reply_params=2)
         return resp[0]
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -772,5 +772,6 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods
 
     def get_remote_data(self, socket_num):
         self._socknum_ll[0][0] = socket_num
-        resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD, self._socknum_ll, reply_params=2)
+        resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD,
+                                self._socknum_ll, reply_params=2)
         return resp[0]
--- a/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
@@ -98,7 +98,7 @@ class WSGIServer:
         """
         self.client_available()
         if (self._client_sock and self._client_sock.available()):
-            self.print_remote_ip()
+            # self.print_remote_ip()
             environ = self._get_environ(self._client_sock)
             result = self.application(environ, self._start_response)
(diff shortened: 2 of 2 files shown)


adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
```diff
@@ -216,3 +230,13 @@ def _get_environ(self, client):
             env[key] = value
 
         return env
+
+    def check_remote_ip(self):
+        sock_num = self._client_sock.socknum
+        remote_ip = _the_interface.get_remote_data(sock_num)
+        return _the_interface.pretty_ip(remote_ip)
+
+    def print_remote_ip(self):
```
jerryneedell: Is it a good idea to have print functions like this in a library? The user can just print the return from check_remote_ip()

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
```diff
@@ -98,9 +98,23 @@ def update_poll(self):
         """
         self.client_available()
         if (self._client_sock and self._client_sock.available()):
+            self.print_remote_ip()
```
jerryneedell: see note below - line 239 - regarding print statements.

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
mytechnotalent: Where is the check_remote_ip() method?

jerryneedell: right above print_remote_ip()

jerryneedell: https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/29900615c81eb9a63407fe87f56f2ec87a37fddd/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py#L234

mytechnotalent: Just making sure you were referring to my method as I was not sure if you were referring to something already existing.

jerryneedell: yes -- I think print statements are normally only used for debug.

mytechnotalent: The check_remote_ip() method is designed to allow forced control if needed otherwise the print method is just for echoing the connections.  Two separate functionalities.

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
mytechnotalent: I will amend the code.

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
jerryneedell: I'll leave it to others to comment. I'm not used to seeing prints in libraries but I'll defer to the other reviewers.

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
jerryneedell: I am not insisting -- just offering an opinion.

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
```diff
@@ -98,7 +98,7 @@ def update_poll(self):
         """
         self.client_available()
         if (self._client_sock and self._client_sock.available()):
-            self.print_remote_ip()
+            # self.print_remote_ip()
```
jerryneedell: sorry -- commenting out code  is not a good idea in my opinion. either remove the function or leave it in.

mytechnotalent:
I don't understand what is causing Travis to fail any ideas?

jerryneedell:
several issues reported here's https://travis-ci.com/adafruit/Adafruit_CircuitPython_ESP32SPI/builds/124470639#L288

mytechnotalent:
I will remove it.

jerryneedell:
Only remove it if you agree with my comments. I may have totally misunderstood your intent.

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
```diff
@@ -95,13 +95,27 @@ def update_poll(self):
...
+
+        Optional routine to control what IP connects to your ESP server
+        as it forces a single connection as the server can't handle more
+        than one request at a time.
+
+        self.client_available()
+        if (self._client_sock and self._client_sock.available()):
+            result = self.check_remote_ip()
+            if result == "192.168.4.2":
+                self.print_remote_ip()
```
jerryneedell: If you have removed the print then this line should also be removed

mytechnotalent: Removed per request.

mytechnotalent: The purpose was to provide new users a usage case so they might be motivated to grow with the platform.

jerryneedell: I had no concern with the example -- only with the line using print_remote_ip() since it was no longer present.

jerryneedell:
Congratulations on satisfying PyLint!!

mytechnotalent:
Thanks @jerryneedell and thank you for the help and suggestions to get the code more functional.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -768,3 +769,12 @@ def get_time(self):
...
         raise RuntimeError("Must be connected to WiFi before obtaining NTP.")
+
+    def get_remote_data(self, socket_num):
+        """
+        Method to obtain remote IP
+        """
+        self._socknum_ll[0][0] = socket_num
+        resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD,
+                                               self._socknum_ll, reply_params=2)
+        return resp[0]
```
mscosti: the `get_remote_data` command in NINA returns 2 reply params. The first is the remote IP, the second is the remote port

instead of just returning the IP, I would return both. 

something like

`return { 'ip_addr': resp[0], 'port': resp[1] }`

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
```diff
@@ -216,3 +217,11 @@ def _get_environ(self, client):
             env[key] = value
 
         return env
+
+    def check_remote_ip(self):
```
mscosti: I would move this to the `Socket` class as a `@property`, like we did for `socknum`.(https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi_socket.py#L183-L186)
We could name the property 'remote_ip`. You could add in a second property for `remote_port` as well.

then every socket instance could get the remote_ip by doing `sock_instance.remote_ip`, and the port by doing `sock_instance.remote_port`

mscosti: if we move the remote ip fetching to a property on `Socket`, then we're still going to want to expose it some way to a server application. Luckily, the WSGI defines an environment request variable named `REMOTE_ADDR`. You could add that to the `get_environ` function here: https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py#L174-L179

it would look something like

`env['REMOTE_ADDR'] = self._client_sock.remote_ip`

The environment object gets passed to the application on every single incoming request, so all your request handlers would just automatically get the remote IP and can use it for processing if they need it.

adafruit_esp32spi/adafruit_esp32spi.py
mytechnotalent: File "/lib/adafruit_esp32spi/adafruit_esp32spi.py", line 550, in pretty_ip
KeyError: 0

mscosti: So by changing the return type an dictionary object means that when you call this method you will need to do an additional step to get the ip whenever you call this method. Dictionary objects have a method on it called `.get(keyname)`. You'll need to update usages of this method so you actually grab the IP address from the dictionary.

Example:
```
remote_data = esp.get_remote_data(socknum)
Ip = remote_data.get('ip_addr')
```

siddacious:
Great work @mytechnotalent , thanks for your contribution!

Thanks to @jerryneedell and @mscosti for the reviews

kattni:
@mytechnotalent @brentru @jerryneedell Thank you for all your work on this PR. It has gotten behind the repo and requires resolving a merge conflict now. I'd like to see this included, but the PR needs to be updated and the final change requests addressed. If you need assistance with this, please let us know. Brent or Jerry: if the original author is unavailable, either of you is welcome to resolve the conflicts and push to this PR.

adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
siddacious: I agree, prints generally don't belong in libraries unless the are behind a debug flag. If you wish to paint information from the library you can expose it with a method or property and display it from your user code

mytechnotalent:
This PR is deprecated in favor of https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/pull/107.

PR #78, opened 2019-10-10
Add booleans for checking if user sets cert/key pair

brentru:
Instead of exposing [`setCert`/`setPSK` in nina-fw](https://github.com/adafruit/nina-fw/blob/master/main/CommandHandler.cpp#L35) as commands, I'm adding two booleans to `esp32spi.py` instead, `set_psk`/`set_crt`.

This is for checking if a user has set a certificate and/or psk prior to establishing a connection with a server.


jerryneedell:
Just a thought -- Is there ever a situation where you would want to clear this setting?

brentru:
@jerryneedell  Possibly if `wifi.connect` with an incorrect cert/key pair and fails.

jerryneedell:
ah, but the user can just do that.
x.set_psk = False

PR #81, opened 2019-10-17
Removed building locally section

dherrada:


dherrada:
@kattni @siddacious

dherrada:
Sounds good

PR #88, opened 2020-01-31
Use secrets.py for ssid and password.

tannewt:


ladyada:
we'd have to update the guide as well

tannewt:
Yup, there are big warnings to manually replace them but it felt weird after setting up the secrets file. I'm curious why it is that way.

ladyada:
used to come before the secrets file, perhaps it got rearranged - @brentru any feelings either way?

dhalbert:
There was a warning below the example; I put another warning above, after helping someone on discord (and being confused myself).

brentru:
@ladyada  I'm OK with this update, it'll reduce confusion.

The Internet Connect! page lists the *`secrets.py` setup* step (https://learn.adafruit.com/adafruit-pyportal/internet-connect#whats-a-secrets-file-17-1) step before the *Connect to WiFi* step.

Once merged and released, I'll change text on the *Connect to WiFi* (https://learn.adafruit.com/adafruit-pyportal/internet-connect#connect-to-wifi-17-4) step to reflect the updated code.

ladyada:
turns out that nobody updated the guide. i did it now.

PR #95, opened 2020-03-25
GAM 20200325 1100 : added optional end of line parameter to readline …

graemorrison:
added optional eol parameter to specify end of line marker if it's not '\r\n'

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -106,11 +106,11 @@ class socket:
         """
         self.send(data)
 
-    def readline(self):
-        """Attempt to return as many bytes as we can up to but not including '\r\n'"""
-        # print("Socket readline")
+    def readline(self,eol=b"\r\n"):
+        """Attempt to return as many bytes as we can up to but not including eol (default='\r\n')"""
+
         stamp = time.monotonic()
-        while b"\r\n" not in self._buffer:
+        while eol not in self._buffer:
             # there's no line already in there, read some more
             avail = self.available()
             if avail:
@@ -118,7 +118,7 @@ class socket:
             elif self._timeout > 0 and time.monotonic() - stamp > self._timeout:
                 self.close()  # Make sure to close socket so that we don't exhaust sockets.
                 raise RuntimeError("Didn't receive full response, failing out")
-        firstline, self._buffer = self._buffer.split(b"\r\n", 1)
+        firstline, self._buffer = self._buffer.split(eol, 1)
         gc.collect()
         return firstline
 
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -106,7 +106,7 @@ class socket:
         """
         self.send(data)
 
-    def readline(self,eol=b"\r\n"):
+    def readline(self, eol=b"\r\n"):
         """Attempt to return as many bytes as we can up to but not including eol (default='\r\n')"""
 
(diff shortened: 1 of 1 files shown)


FoamyGuy:
@graemorrison Thank you for this change! I will work on testing it tonight and we should be able to get it merged in soon.

graemorrison:
it's needed because not all systems terminate a line with '\r\n' (such as unix), in fact if you look at  your example in https://docs.python.org/3/library/io.html#io.IOBase.readline it actually specifies '\n' as the eol terminator and it's being modified in your version to be '\r\n'

FoamyGuy:
@tannewt I think IOBase `readline()` is different because it's reading from the terminal. This is for socket `readline()` If I am understanding correctly it seems that the CPython socket module does not provide a `readline()` function at all, the closest things seems to be makefile: https://docs.python.org/3/library/socket.html#socket.socket.makefile which would allow for iterating over the received data one line at a time I think. makefile() does have a `newline`, though I don't see anywhere in the documentation that states specifically what it does, I would assume it determines the characters used within the file like object that it returns. I'm definitely no expert though perhaps I am misunderstanding something.

graemorrison:
@tannewt - sorry, my previous response was little curt. Essentially the reason for the kwarg is to provide backwards compatibility with the existing version, hence the default being the current setting.

tannewt:
> it's needed because not all systems terminate a line with '\r\n' (such as unix), in fact if you look at your example in https://docs.python.org/3/library/io.html#io.IOBase.readline it actually specifies '\n' as the eol terminator and it's being modified in your version to be '\r\n'

I understand this but don't know why it applies to ESP32SPI which is reading from the ESP only. Does the ESP pass line endings as returned by different servers?

My desire in the long-term is to remove this function because it doesn't match CPython's socket API (as @FoamyGuy points out). I'd much prefer that our socket classes are strict subset so that we can test things that build on top of it with CPython. I'll likely do this as we add Wifi support for the ESP32-S2.

So, I'd rather not change this API since it will be removed later and shouldn't be used in more places. Instead, I'd encourage you to use the `recv*_into` methods from socket instead (and add them if they are missing). They are much *much* better since they do not allocate new buffers at all.

graemorrison:
> > it's needed because not all systems terminate a line with '\r\n' (such as unix), in fact if you look at your example in https://docs.python.org/3/library/io.html#io.IOBase.readline it actually specifies '\n' as the eol terminator and it's being modified in your version to be '\r\n'
> 
> I understand this but don't know why it applies to ESP32SPI which is reading from the ESP only. Does the ESP pass line endings as returned by different servers?
> 
> My desire in the long-term is to remove this function because it doesn't match CPython's socket API (as @FoamyGuy points out). I'd much prefer that our socket classes are strict subset so that we can test things that build on top of it with CPython. I'll likely do this as we add Wifi support for the ESP32-S2.
> 
> So, I'd rather not change this API since it will be removed later and shouldn't be used in more places. Instead, I'd encourage you to use the `recv*_into` methods from socket instead (and add them if they are missing). They are much _much_ better since they do not allocate new buffers at all.

@tannewt, yes the ESP is ignorant of line endings it's simply returning the data as transmitted by the remote end of a network session, depending on the application running at the other side it could be anything - <cr><nl>, <nl>, could even be <fs> or <rs> or just about any other kind of delimiter.

However I get your point about compatibility and not diverging from the upstream package, especially if the goal is to remove this in a near-future release.
I will look at the 'recv' routines and see what's appropriate to what I'm trying to do 
thanks for the advice - I'll remove this PR

tannewt:
Thanks for understanding @graemorrison ! Let me know if I can help with the recv stuff!

Issue #63, opened 2019-08-13
Connection Sleep or Timeout


calittle:
This is an enhancement request to provide for dropping a connection and shutting off the radio after a specified time period of no use or after a fixed time period. Upon request e.g. post(), WifiManager will turn on the radio and reestablish the connection and reset the timer. This could be useful for battery-powered projects where always-on radio and connections will drain the battery quickly.

siddacious:
@brentru Does the NINA firmware have a shutdown feature?

mscosti:
I'm not sure exactly what 'low power mode' does, but NINA does support putting it into a low power mode: https://github.com/adafruit/nina-fw/blob/master/main/CommandHandler.cpp#L148-L162

ladyada:
thats as good as yr gonna get. anyone want to implement that function as a PR? :)

anecdata:
The two pull requests above summarize the current state of power control of the ESP32 NINA co-processor:
- NINA sets a nominal lower power mode by default that reduces power usage about roughly half when not communicating with the AP. This is the normal level of opereation already. There is a higher power saving mode that would require NINA and ESP32SPI changes, but also may introduce operational differences.
- A `disconnect()` function is now available in `ESP_SPIcontrol`, reducing average power usage to about half in common use cases. This function can now be used directly in user code, or could be incorporated into higher-level libraries like WiFiManager.

brentru:
> A disconnect() function is now available in ESP_SPIcontrol, reducing average power usage to about half in common use cases. This function can now be used directly in user code, or could be incorporated into higher-level libraries like WiFiManager.
Exciting!

anecdata:
Just a note that ESP32SPI NTP using `esp.get_time()` doesn't play well with staying disconnected until needed since the ESP32 needs some interval (~15 seconds I believe) after connection before it will return valid seconds.

PR #101, opened 2020-06-06
Wifi low power

anecdata:
Expose the NINA firmware `setPowerMode` function. Note that:

> Modem-sleep mode works in Station-only mode and the station must connect to the AP first.

NINA firmware currently implements only `WiFi.lowPowerMode()` or `WiFi.noLowPowerMode()`, and low power mode defaults to `WIFI_PS_MIN_MODEM`: 

> Minimum modem power saving. In this mode, station wakes up to receive beacon every DTIM period

Preliminary testing of low power mode vs. no low power mode indicates potentially significant difference in power usage, but it is dependent on AP settings and wi-fi usage within the client CircuitPython application.

However, though I may be misunderstanding something, it appears that the default NINA build enables `WIFI_PS_MIN_MODEM` (though I don't pretend to understand the build process and can't find where it's done in the code). Crude testing with this PR code shows similar battery life for code with no call to `esp.wifi_set_power_mode()` and call to `esp.wifi_set_power_mode(True)`. Call to `esp.wifi_set_power_mode(False)` does increase power usage, but there may be a benefit in some cases:

> provides minimum latency for receiving Wi-Fi data in real time

Background information is here:
[https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/wifi.html#esp32-wi-fi-power-saving-mode](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/wifi.html#esp32-wi-fi-power-saving-mode)

A future enhancement for lower power usage, which would require a change to NINA, could add the option for the higher power saving mode by using `esp_wifi_set_ps()` with the `WIFI_PS_MAX_MODEM` value (caveats in the link above), instead of the current call to`WiFi.lowPowerMode()`.

There are other low-power features of the ESP32, but they would require development from scratch in the NINA firmware, plus exposure in this library. There's always the `EN` pin for super low power usage ;-)

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -451,7 +451,8 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
 
     def wifi_set_power_mode(self, power_mode):
         """Sets wi-fi power save mode on or off"""
-        resp = self._send_command_get_response(_SET_POWER_MODE_CMD, ((bool(power_mode),),))
+        resp = self._send_command_get_response(_SET_POWER_MODE_CMD,
+                                               ((bool(power_mode),),))
         if resp[0][0] != 1:
             raise RuntimeError("Failed to set power mode")
 
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -451,8 +451,9 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
 
     def wifi_set_power_mode(self, power_mode):
         """Sets wi-fi power save mode on or off"""
-        resp = self._send_command_get_response(_SET_POWER_MODE_CMD,
-                                               ((bool(power_mode),),))
+        resp = self._send_command_get_response(
+            _SET_POWER_MODE_CMD, ((bool(power_mode),),)
+        )
         if resp[0][0] != 1:
             raise RuntimeError("Failed to set power mode")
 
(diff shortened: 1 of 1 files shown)


anecdata:
Relevant to issue #63 

I'd welcome some eyes on this, especially to confirm that NINA already uses MIN modem low power mode. I have an INA219 on order to make better measurements.

In any case, this would allow the option of turning modem low power mode on and off.

jerryneedell:
@anecdata does this require any specific build of the NiNa Firmware?

anecdata:
@jerryneedell No, it's independent of any other feature added, and should work all the way back to 1.2.2 at least.

The only caveat I'd add is if the build or IDF changed the default modem low power mode at some point, then default behavior could be different.

jerryneedell:
better example -- just run esp32spi_simpletest
 with normal settings -- after completions current goes back to ~70mA

with wiifi_set_power_mode(False) it stays at ~140mA after completion

anecdata:
Thanks, Jerry. That's similar to the pattern I observed, and appears to confirm that MIN power save is on by default (at least in recent version of NINA).

anecdata:
From research for this issue, it seemed the feature in IDF pre-dates our NINA (2017-ish), although there was a fix in IDF 3.2.2 (which I'm not sure of, but we may not have picked up until NINA 1.4.0) for buffering broadcast/multicast during modem low power mode.

anecdata:
@brentru Any thoughts?

If there's little value in being able to turn _off_ low-power (or perhaps that and using this as a stub for later adding the `WIFI_PS_MAX_MODEM`higher-power-save mode (which needs NINA changes), then we can just close this.

brentru:
@anecdata  Agreed - let's close this in favor of adding the more effective power save mode in nina-fw. Thanks for trying it out.

PR #104, opened 2020-08-23
remove bad-whitespace pylint directive

jepler:
This directive has been removed from pylint 2.6.0.


jepler:
Fails due to unrelated pylint wrong-import-order errors I do not reproduce locally with pylint 2.6.0, such as
```
adafruit_esp32spi/adafruit_esp32spi_wifimanager.py:20:0: C0411: third party import "import adafruit_requests as requests" should be placed before "from adafruit_esp32spi import adafruit_esp32spi" (wrong-import-order)
```

kattni:
@jepler I would suggest updating the import order, but be prepared for it to cascade fail on the list of imports one at a time. Your local environment probably isn't _exactly_ like the remote environment, and that's why it's not failing locally for you. I've had it happen.

PR #106, opened 2020-09-28
adding optional service function

siddacious:
I got frustrated by how much time my sensor project was taking trying to talk to IO, so I did some digging and found that the vast majority of the time taken to do a POST to IO (or probably any HTTP request) is spent polling the ready pin of the Airlift module to see if it's ready to talk.

I added an optional argument to the ESP32SPI class to allow the user to provide a function that gets called each cycle while waiting  on the ESP32. Super useful! Now I can read buttons and do stuff while waiting for WiFi to be done

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ bundles
 .eggs
 dist
 **/*.egg-info
+.vscode/
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -134,8 +134,17 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
 
     # pylint: disable=too-many-arguments
     def __init__(
-        self, spi, cs_pin, ready_pin, reset_pin, gpio0_pin=None, *, debug=False
+        self,
+        spi,
+        cs_pin,
+        ready_pin,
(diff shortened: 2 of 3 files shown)


tannewt:
Is it waiting for ready in the middle of a request or at the start? If it's the start, then we could expose readiness directly. Passing in a function to call when idle is definitely advanced and needs documentation at the class level in addition to the example.

siddacious:
@tannewt it's waiting whenever it wants to send a command to the ESP which happen happens nearly every time you interact with it

tannewt:
I'm wary of adding this callback because it's a non-standard way to do concurrency AND this library is a huge hodge podge already.

Instead, we could actually start to do asyncio per the long CircuitPython issue: https://github.com/adafruit/circuitpython/issues/1380

For monitoring buttons, could you use gamepad in the meantime?

siddacious:
It is indeed a bit of a hack, but it was easier than refactoring the whole library to have a non-blocking API 😅 

I am already using gamepadshift to read the buttons on an edgebadge but I can't call it to check for button presses if the ESP32SPI library has the only thread/instruction pointer/whatever.

I'm all for starting on async/concurrency. I've been doing a lot more "eventy" coding recently and I fear that if not given additional tools I might accidentally write full-fledged OS :|

I'll close this for now but I'm going to open an issue for the context thievery if there isn't already one

PR #107, opened 2020-09-29
Reorg code per request from orphan pr 69

mytechnotalent:
This PR deprecates PR 69 as a year has passed as I was tied up with other responsibilities.  The code base has changed significantly so there was a large merge conflict with the prior work.  This new PR takes all of the recommendations of @mscosti @brentru @jerryneedell into account.  @kattni apology for taking so long.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -66,6 +66,7 @@ _GET_HOST_BY_NAME_CMD = const(0x35)
 _START_SCAN_NETWORKS = const(0x36)
 _GET_FW_VERSION_CMD = const(0x37)
 _SEND_UDP_DATA_CMD = const(0x39)
+_GET_REMOTE_DATA_CMD = const(0x3A)
 _GET_TIME = const(0x3B)
 _GET_IDX_BSSID_CMD = const(0x3C)
 _GET_IDX_CHAN_CMD = const(0x3D)
@@ -895,6 +896,15 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
                 "Cannot obtain NTP while in AP mode, must be connected to internet"
             )
         raise RuntimeError("Must be connected to WiFi before obtaining NTP.")
+
+    def get_remote_data(self, socknum):
+        """Obtain remote IP
+        :param int socknum: the socket number.
+        """
+        self._socknum_ll[0][0] = socknum
+        resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD,
+                                               self._socknum_ll, reply_params=2)
+        return { 'ip_addr': resp[0], 'port': resp[1] }
 
     def set_certificate(self, client_certificate):
         """Sets client certificate. Must be called
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -317,7 +317,7 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
         *,
         reply_params=1,
         sent_param_len_16=False,
-        recv_param_len_16=False
+        recv_param_len_16=False,
     ):
         """Send a high level SPI command, wait and return the response"""
         self._send_command(cmd, params, param_len_16=sent_param_len_16)
@@ -902,9 +902,10 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
         :param int socknum: the socket number.
         """
         self._socknum_ll[0][0] = socknum
-        resp = self._send_command_get_response(_GET_REMOTE_DATA_CMD,
-                                               self._socknum_ll, reply_params=2)
-        return { 'ip_addr': resp[0], 'port': resp[1] }
+        resp = self._send_command_get_response(
+            _GET_REMOTE_DATA_CMD, self._socknum_ll, reply_params=2
+        )
+        return {"ip_addr": resp[0], "port": resp[1]}
 
     def set_certificate(self, client_certificate):
         """Sets client certificate. Must be called
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -193,6 +193,16 @@ class socket:
         """The socket number"""
         return self._socknum
 
+    @property
+    def remote_ip(self):
+        """The remote ip"""
+        return self._remote_ip
+
+    @property
+    def remote_port(self):
+        """The remote port"""
+        return self._remote_port
+
     def close(self):
         """Close the socket, after reading whatever remains"""
         _the_interface.socket_close(self._socknum)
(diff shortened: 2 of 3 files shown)


tannewt (changes requested):
Thanks for updating these changes!

Please only add CPython compatible APIs to socket. Looks like you want `getpeername`: https://docs.python.org/3/library/socket.html#socket.socket.getpeername

You cannot add it to the constructor because CPython doesn't have it.

mytechnotalent:
@tannewt there is no `getpeername` exposed in the Nina firmware https://github.com/adafruit/nina-fw/blob/master/main/CommandHandler.cpp.  The `getpeername` from CPython exists here https://github.com/python/cpython/blob/e822e37946f27c09953bb5733acf3b07c2db690f/Modules/socketmodule.c#L3377.

Trying to add this API would be a major amount of work and to be honest without it following the convention of the rest of this lib I would not even know where to begin.  Do you think this PR should simply be closed if it might not be a feature worth implementing?

tannewt:
@mytechnotalent I don't think you need getpeername exposed from Nina because you already have a way to get the remote ip. I think you can call that API from your `getpeername` implementation.

mytechnotalent:
Can you provide move information?  Perhaps an example?

adafruit_esp32spi/adafruit_esp32spi_socket.py
```diff
@@ -52,14 +52,23 @@ class socket:
...
+        remote_port=None,
     ):
         if family != AF_INET:
             raise RuntimeError("Only AF_INET family supported")
         if type != SOCK_STREAM:
             raise RuntimeError("Only SOCK_STREAM type supported")
         self._buffer = b""
         self._socknum = socknum if socknum else _the_interface.get_socket()
+        self._remote_ip = remote_ip
+        self._remote_port = remote_port
```
tannewt, line 71:
```suggestion
```

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -896,6 +897,16 @@ def get_time(self):
...
 
+    def get_remote_data(self, socknum):
+        """Obtain remote IP
+        :param int socknum: the socket number.
+        """
+        self._socknum_ll[0][0] = socknum
+        resp = self._send_command_get_response(
+            _GET_REMOTE_DATA_CMD, self._socknum_ll, reply_params=2
+        )
+        return {"ip_addr": resp[0], "port": resp[1]}
```
tannewt, line 908:
```suggestion
        return (resp[0], resp[1])
```

adafruit_esp32spi/adafruit_esp32spi_socket.py
```diff
@@ -193,6 +202,16 @@ def socknum(self):
...
+    @property
+    def remote_ip(self):
+        """The remote ip"""
+        return self._remote_ip
+
+    @property
+    def remote_port(self):
+        """The remote port"""
+        return self._remote_port
+
```
tannewt, line 214:
```suggestion
    def getpeername(self):
        ip, port = _the_interface.get_remote_data(self._socknum)
        # Parse so that ip is a string and port is an int
        # Second bullet here: https://docs.python.org/3/library/socket.html#socket-families
        return (ip, port)

```

adafruit_esp32spi/adafruit_esp32spi_socket.py
```diff
@@ -52,14 +52,23 @@ class socket:
...
     def __init__(
-        self, family=AF_INET, type=SOCK_STREAM, proto=0, fileno=None, socknum=None
+        self,
+        family=AF_INET,
+        type=SOCK_STREAM,
+        proto=0,
+        fileno=None,
+        socknum=None,
+        remote_ip=None,
+        remote_port=None,
```
tannewt, line 62:
```suggestion
```

tannewt (changes requested):
Here is a sketch of the suggested changes. I haven't tested it. My goal was to give an example of what I'd like to see.

adafruit_esp32spi/adafruit_esp32spi_socket.py
mytechnotalent: Oh that makes sense @tannewt!  I will work on these changes as soon as I get a second and will test and push an update.  Thank you again.

tannewt: No rush and no problem! Thanks!

mytechnotalent: @tannewt I have implemented the changes.  I wanted ask specifically how you would like this tested.  Here is the result in the  REPL.
`>>> socket.socket.
close           connect         connected       read
readline        recv            send            settimeout
write           socknum         available       getpeername`

I obviously cant call `socket.socket.getpeername()` so if you have a quick test I can code up I will begin testing so that I know what you are looking for specifically.  Any simple example will do thanks!

tannewt: I'd imagine you need to create a socket and connect before calling it. The new requests library can work with CPython sockets so that's probably the best place to start.

mytechnotalent: @tannewt I created an access point called `cool` and then I connected my iPhone to it.  It successfully connected however when I call `socket.socket.getpeername()` i get `TypeError: function takes 1 positional arguments but 0 were given` as it should be returning the peer ip of `192.168.4.2` as that is the iPhone and the server is `192.168.4.2`. 

Any suggestions on how to better check the `getpeername` call?

tannewt: You actually need to connect a TCP socket and then call .getpeername() on it. The error is because you are calling it on the class, not an instance, and therefore no `self` is passed.

Why are you adding this overall? It doesn't seem like you have a use for it.

mytechnotalent: Great question @tannewt.  Over a year ago when I did the original PR I had issues where I had the device set up as an Access Point and if I accidently logged in from multiple browsers on various machines it would crash. 

This original PR was designed to check for that remote IP and if it was not 192.168.4.2 then don't connect.  It was to add the ability to get this information if a developer wanted to implement functionality such as this.

I am ok with closing the PR however.  Thoughts?

mytechnotalent: This is what I tried to do originally as well.  I can't seem to even create a socket object here.
`>>> import adafruit_esp32spi.adafruit_esp32spi_socket as socket
>>> s = socket.socket()
  File "<stdin>", line 1, in <module>
  File "/lib/adafruit_esp32spi/adafruit_esp32spi_socket.py", line 62, in __init__
AttributeError: 'NoneType' object has no attribute 'get_socket'
>>> `

tannewt: With the current ESP32-SPI implementation you need to call `set_interface` on the socket module: https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi_socket.py#L24 I believe you pass it the ESP32SPI object.

I could see this PR being helpful but if no one is using it enough to test, then we could just close for now.

Issue #110, opened 2020-10-24
TypeError when calling ESPSPI_WiFiManager.signal_strength


mikejc58:
In adafruit_esp32spi_wifimanager.py, ESPSPI_WiFiManager.signal_strength() accesses self.esp.rssi as a function, but it is a property.

`return self.esp.rssi()`  should be `return self.esp.rssi`

siddacious:
Fixed via #111

PR #111, opened 2020-10-26
Rssi fix for issue reported in #110

siddacious:


ladyada (approved):

ladyada:
a timer example was also added?

siddacious:
@ladyada looks like my stash failed! I'll fix

PR #115, opened 2020-11-12
wsgiserver: implement parse_headers

Anton-2:
This PR implements parse_headers, which is not available in adafruit_requests anymore.

It also adds a default return status of 500 Internal Server Error, this helped me when debugging an app using Adafruit_CircuitPython_WSGI.

Lastly, we always add a Connection: close header. The server is not ready to handle keep-alive requests, it seems better to make it clear for the client.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
@@ -52,7 +52,7 @@ def parse_headers(client):
         line = str(client.readline(), "utf-8")
         if not line:
             break
-        title, content = line.split(':', 1)
+        title, content = line.split(":", 1)
         headers[title.strip().lower()] = content.strip()
     return headers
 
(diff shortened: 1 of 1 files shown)


astrilchuk:
Thanks @Anton-2 for this fix.  I hope it gets merged soon since the wsgiserver is currently broken.

siddacious (changes requested):
@Anton-2 Apologies for the delay; `parse_headers` exists in request as `_parse_headers` a private-ish method of the class, WSGIServer needs to be updated to not try to parse the headers itself and instead access the parsed headers with `Response.headers`.

jposada202020:
@Anton-2 this is a long time (I know) are you still interested on working on this, if so, could you take intro consideration the comments above.  let us know thanks :)

Neradoc:
`wsgiserver` does not make requests, it only uses the `parse_headers` function to pass them to the application. I might be missing something, but I don't see how or why we would want to use the `Response` class here instead of just duplicating the function from the old requests (and drop the dependency) like the PR does.

jposada202020:
@brentru Hello :) what are your thoughts on this?
Thanks

brentru:
The proposed changes make sense to me.

jposada202020:
Good Thanks, if that is the case I think this is settled, I will let some day to merge to see after bumping this PR other folks have comments. Thanks

PR #117, opened 2020-11-17
allow for list in secrets[ssid]

slootsky:
check secrets["ssid"] to see if it is a list or a tuple, and if so, loop through it

adafruit_esp32spi_wifimanager allows a list or tuple, and this brings adafruit_esp32spi into line with that.

This code gets the ESP to use a list or a string for secrets["ssid"]

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -31,7 +31,7 @@ from micropython import const
 from digitalio import Direction
 from adafruit_bus_device.spi_device import SPIDevice
 
-__version__ = "0.0.0-auto.0"
+__version__ = "3.5.3"
 __repo__ = "https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI.git"
 
 _SET_NET_CMD = const(0x10)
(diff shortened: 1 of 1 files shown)


adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -31,7 +31,7 @@
 from digitalio import Direction
 from adafruit_bus_device.spi_device import SPIDevice
 
-__version__ = "0.0.0-auto.0"
+__version__ = "3.5.3"
```
tannewt, line 34: Please undo this. The original version is automatically replaced.

tannewt (changes requested):
I don't think this belongs in this library. It should be up to the user code to decide what essid to connect to.

adafruit_esp32spi/adafruit_esp32spi.py
slootsky: no problem undoing that.  FYI:  pre-commit added it.

slootsky:
a) that test failure doesn't have anything to do with my code, but I don't know how to force a retest without making an unnecessary code change.
b) with regards to your comment that this doesn't belong in this library, the adafruit_esp32spi_wifimanager accepts ssid and password as lists/tuples.  This makes this code more compatible with wifimanager.  If the user wants to supply a string, then they get explicit control over which ssid they connect to, otherwise the connection attempts happen in the order of the list in secrets["ssid"]

tannewt:
Why can't you use wifimanager then? This seems like a very special case for this class.

slootsky:
this code makes the ssid and password in the secrets dictionary compatible between wifimanager and this class.  Without this, a list that works in wifimanager won't work here.

Having multiple ssids configured is good for a device that may not be in the same location all the time

tannewt:
I understand that this code makes them match. What is not clear to me is why you aren't just using WifiManager already. This code already does way more than it should. There is no reason this low level needs to manage what essid to connect to.

slootsky:
Sounds like you don't think this code should be modified to accept a list.

Would you like me to close the pull request then?

tannewt:
@slootsky Yup. You could provide an example or just close. Thank you for the PR anyway!

PR #118, opened 2020-11-22
Add support for Bluepad32 gamepads

ricardoquesada:
This commit adds a new message to the SPI protocol: get_gamepad_data()

This new function uses the unused 0x60 message ID and requires the
Bluepad32 firmware, since the Nina-fw firmware doesn't support it yet.

More information here:
https://gitlab.com/ricardoquesada/bluepad32/-/blob/master/docs/plat_airlift.md

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -87,6 +87,8 @@ _SET_ANALOG_WRITE_CMD = const(0x52)
 _SET_DIGITAL_READ_CMD = const(0x53)
 _SET_ANALOG_READ_CMD = const(0x54)
 
+_GET_GAMEPADS_DATA = const(0x60)
+
 _START_CMD = const(0xE0)
 _END_CMD = const(0xEE)
 _ERR_CMD = const(0xEF)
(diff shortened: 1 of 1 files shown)


dherrada:
Hey @ricardoquesada could you run black and then commit those changes? Here are the instructions for doing that: https://learn.adafruit.com/improve-your-code-with-pylint/black

ricardoquesada:
> Hey @ricardoquesada could you run black and then commit those changes? Here are the instructions for doing that: https://learn.adafruit.com/improve-your-code-with-pylint/black

Thanks. There were a few warnings. I fixed them, updated pull-request and it is still failing.

What I'm doing is the following:

```sh
$ cd ~/src/Adafruit_CircuitPython_ESP32SPI
$ pylint adafruit_esp32spi/adafruit_esp32spi.py 

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)
```

Any idea on what I'm missing? Thanks!

FoamyGuy:
@ricardoquesada it's wanting the Black code formatter to be run on the code. If you already ran it once, then some of the subsequent changes may have caused it to find a new thing it wants formatted. 

If you run the Black formatter on `adafruit_esp32spi.py` I think the current error will be resolved.

ricardoquesada:
> @ricardoquesada it's wanting the Black code formatter to be run on the code. If you already ran it once, then some of the subsequent changes may have caused it to find a new thing it wants formatted.
> 
> If you run the Black formatter on `adafruit_esp32spi.py` I think the current error will be resolved.

my bad... I was running `pylint` instead of `black`.

FoamyGuy:
No worries. We use both so fixing up the pylint stuff is good as well, it would complain about that next ;-) 

Thank you for working through all of the CI issues.

tannewt (changes requested):
I don't want to add a non-nina feature to this library. Ideally we wouldn't have anything but wifi related things in here.

You can either fork this library for yourself or create a separate library that takes in an ESP32SPI object and uses it. It shouldn't be done in this library.

ricardoquesada:
> ...create a separate library that takes in an ESP32SPI object and uses it. It shouldn't be done in this library.

sounds reasonable to me.

PR #122, opened 2021-01-05
Expose NINA disconnect

anecdata:
Expose the existing `disconnect` [function](https://github.com/adafruit/nina-fw/blob/f2a0e601b23212dda4fe305eab30af49a7c7fb41/main/CommandHandler.cpp#L680) in NINA, so that a CircuitPython client can explicitly disconnect from the connected AP.

There are a variety of use cases for this, including saving power when Wi-Fi is not needed (see [https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/issues/63](https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/issues/63) and [https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/pull/101](https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/pull/101)).

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -505,6 +505,12 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
             self.reset()
             return False
 
+    def disconnect(self):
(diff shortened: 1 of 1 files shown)


tannewt (approved):
Thanks!

PR #125, opened 2021-03-05
Add usage example for RP2040

brentru:
Adding new simpletest example for RP2040 pico to include inline learn.


brentru:
NOTE: This is a separate simpletest example due to the non-traditional SPI bus constructor (not `board.` objects) and for a new guide.

ladyada (approved):

PR #126, opened 2021-03-15
Create esp32spi_tcp_client.py

anecdata:
A simple TCP client example using ESP32SPI.

Requires running a TCP server listening on the same port. For example, CPython on a general-purpose computer:
https://github.com/adafruit/circuitpython/blob/main/tests/circuitpython-manual/socketpool/client/host-server.py
or CircuitPython native sockets:
https://github.com/adafruit/circuitpython/blob/main/tests/circuitpython-manual/socketpool/server/cpy-server.py

### DIFF
--- a/examples/esp32spi_tcp_client.py
+++ b/examples/esp32spi_tcp_client.py
@@ -37,7 +37,7 @@ print("Connecting")
 s.connect(socketaddr)
 
 print("Sending")
-s.send(b'Hello, world')
+s.send(b"Hello, world")
 
 print("Receiving")
 print(s.recv(128))
--- a/examples/esp32spi_tcp_client.py
+++ b/examples/esp32spi_tcp_client.py
@@ -1,0 +1,43 @@
+# SPDX-FileCopyrightText: 2021 Adafruit Industries
+# SPDX-License-Identifier: MIT
+
+import board
+from digitalio import DigitalInOut
+from adafruit_esp32spi import adafruit_esp32spi
+import adafruit_esp32spi.adafruit_esp32spi_socket as socket
+import adafruit_requests as requests
+from secrets import secrets
+
+
+TIMEOUT = 5
+# adjust host and port to match server
+HOST = "192.168.10.179"
+PORT = 5000
+
+# PyPortal or similar; adjust pins as needed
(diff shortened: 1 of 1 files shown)


examples/esp32spi_tcp_client.py
```diff
@@ -0,0 +1,43 @@
...
+from digitalio import DigitalInOut
+from adafruit_esp32spi import adafruit_esp32spi
+import adafruit_esp32spi.adafruit_esp32spi_socket as socket
+import adafruit_requests as requests
+from secrets import secrets
+
+
+TIMEOUT = 5
+# adjust host and port to match server
+HOST = "192.168.10.179"
```
brentru, line 14: Might want to add a different IP or placeholder text/IP here?

examples/esp32spi_tcp_client.py
```diff
@@ -0,0 +1,43 @@
...
+PORT = 5000
+
+# PyPortal or similar; adjust pins as needed
+spi = board.SPI()
```
brentru, line 18: Same here, or pull from secrets

examples/esp32spi_tcp_client.py
anecdata: different IP? I'm happy to dummy it up somehow, not sure what you're thinking?

examples/esp32spi_tcp_client.py
anecdata: I agree secrets would be a good approach in general, but it isn't standard for host and port, so would require explanatory text on how to do that... so probably same approach as above

examples/esp32spi_tcp_client.py
anecdata: I could use wifitest.adafruit.com and port 80 ...that would give some 404 rexponse or similar

anecdata: `Receiving
b'HTTP/1.1 400 Bad Request\r\nServer: nginx/1.10.3 (Ubuntu)\r\nDate: Mon, 15 Mar 2021 16:24:42 GMT\r\nContent-Type: text/html\r\nContent-L'`

brentru (approved):

PR #127, opened 2021-03-18
UDP Client using socket

anecdata:
UDP Client was previously possible using `esp` calls directly. This change allows UDP Client at the `socket` level.

PR #128, opened 2021-03-18
UDP Client example (NTP)

anecdata:
using socket API rather than esp API


jposada202020 (commented):
#### Looks good to me

Reviewed with new changes in module and works as expected for tests and results see PR #127

anecdata:
@jposada202020 Thanks for testing!

jposada202020:
@anecdata my pleasure

Issue #121, opened 2020-12-28
__init__ arguments are not Pins


dhalbert:
The arg names here are suffixed as `_pin`, but they should actually be `DigitalInOut` objects. The arg names could be clearer, maybe `_dio` or just drop the suffix. At least one user might have been confused by this: https://forums.adafruit.com/viewtopic.php?f=60&t=173353.

https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/fce466bd2bb70ca86b79e5cb36bbaca00afacfd1/adafruit_esp32spi/adafruit_esp32spi.py#L136-L138

brentru:
@dhalbert  I think dropping the suffix would be clearer compared to substituting `_pin` for  `_dio`

tekktrik:
Is this something we want to do?  I'm happy to submit a PR for this :) Was also planning to add type annotations, if that helps at all.

FoamyGuy:
resolved by #146

FoamyGuy:
@tekktrik all of the other instances of similar `*_pin` named arguments are linked above here in this issue now.

tannewt:
I agree with @brentru that dropping the suffix is clearer than changing it to `_dio`. Adding the suffix is redundant with the type info.

tekktrik:
I can submit another PR to drop it if that's the case

tannewt:
I think it's important to point out that by taking DigitalInOut its usually meant to work with GPIO expander provided objects as well. Pin should be only used when a native pin object is expected.

PR #146, opened 2021-12-29
Change _pin arguments to _dio

tekktrik:
Addresses Issue #121 by changing relevant argument names from `*_pin` to `*_dio`.  I chose this instead of just using `*` argument names so it would be explicitly clear it should be a `DigitalInOut` instance passed as the argument, and also because I know there is a `reset` parameter so that argument wouldn't be confused for a boolean argument or something like that.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -134,7 +134,7 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
 
     # pylint: disable=too-many-arguments
     def __init__(
-        self, spi, cs_pin, ready_pin, reset_pin, gpio0_pin=None, *, debug=False
+        self, spi, cs_dio, ready_dio, reset_dio, gpio0_dio=None, *, debug=False
     ):
         self._debug = debug
         self.set_psk = False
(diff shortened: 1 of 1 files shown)


ladyada:
can you check if any examples/code use the old naming?

tekktrik:
Will do!

tekktrik:
It looks like all the pin arguments in the examples in the repo are named `esp32_*` like `esp32_reset`.  Should I check the Learn guides as well, or would those reference the examples?

ladyada:
yah check the learn system, its kinda risky to change arg names :/ lots of code can break

tekktrik:
A quick search for ESP32 products and reading the guides shows that the examples (no references other than code examples) always use the same variable naming scheme of `esp32_*` and the arguments are never done as keywords, only positional.

tekktrik:
Is this all good?  Happy to do anything needed if this is a viable PR still!

ladyada:
@jerryneedell any thoughts? just cause you deal with this stuff more recently than i!

jerryneedell:
No concerns from me. Than you for asking!

anecdata (commented):
I've only seen instantiation positional, typically with variables assigned from the `DigitalInOut`s:
`esp = adafruit_esp32spi.ESP_SPIcontrol(spi, esp32_cs, esp32_ready, esp32_reset)`
Tested PR code on PyPortal, change is transparent.

tekktrik:
Any decision on this? Happy to change the names to something else as well if those aren't clear.

FoamyGuy (approved):
Looks good to me, I agree in thinking that `*_dio` makes it more clear what is expected to be passed for those arguments.

I tested these changes successfully with the requests simpletest on a PyPortal 7.2.0-alpha.1.

I also used `git grep "cs_pin="` and similar to search for usages of these argument names in the Learning System repo and did not find any code using those names to instantiate this library.  

However there are a few other libraries with similarly named `*_pin` arguments that are actually expecting DigitalIO objects like this one. I'll open issues on those repo's I would assume we want to be consistent about the argument naming across repos.

tekktrik:
Sounds good!  I'd be happy to look into how easily they can get a PR without breaking code!

Issue #120, opened 2020-12-26
Documentation on status LED?


ajs256:
I can't seem to find documentation on how this library controls the NeoPixel on my MatrixPortal. Could someone please point me to where I could find that?

jerryneedell:
This is probably not what you were asking for but hopefully it will help.
It is used by the Wifi manager  https://learn.adafruit.com/adafruit-pyportal/internet-connect#wifi-manager-3034791-36 , but you are correct, the actual usage is not documented. The simplest thing now would be to look at the code for places where the pixel_status is set -- for example 
https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py#L139
and
https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/master/adafruit_esp32spi/adafruit_esp32spi_wifimanager.py#L142

show it being set red  (100,0,0) before then green(0,100,0) after "connecting".

FoamyGuy:
closed by #148

PR #148, opened 2021-12-30
Add documentation for status LED

tekktrik:
Added documentation for the behavior of the status LED in the `__init__()` parameter to address Issue #120.  I went with this way instead of for each method for brevity, but if that would be the preferred method, I'm happy to do it that way!

Also changed it so `post()` would turn the status LED off after completion like the other HTTP methods.


FoamyGuy (approved):
Looks good to me. Thanks @tekktrik

Issue #154, opened 2022-01-22
recv_into() missing second argument


jasonwashburn:
Any time I try to use the requests library on my Pyportal since upgrading to CircuitPython 7.1.1 I get the following error:
```
code.py output:
Connecting to AP...
Connected to WBurn-2G 	RSSI: -50
Fetching text from http://wifitest.adafruit.com/testwifi/index.html
Traceback (most recent call last):
  File "code.py", line 58, in <module>
  File "adafruit_requests.py", line 847, in get
  File "adafruit_requests.py", line 719, in request
  File "adafruit_requests.py", line 210, in __init__
  File "adafruit_requests.py", line 297, in _readto
  File "adafruit_requests.py", line 241, in _recv_into
TypeError: function takes 2 positional arguments but 3 were given

Code done running.
```

I've tried multiple example files that use the library with the same result. The currently loaded version is  the code below.
```
# SPDX-FileCopyrightText: 2021 ladyada for Adafruit Industries
# SPDX-License-Identifier: MIT

# adafruit_requests usage with an esp32spi_socket
import board
import busio
from digitalio import DigitalInOut
import adafruit_esp32spi.adafruit_esp32spi_socket as socket
from adafruit_esp32spi import adafruit_esp32spi
import adafruit_requests as requests

# Add a secrets.py to your filesystem that has a dictionary called secrets with "ssid" and
# "password" keys with your WiFi credentials. DO NOT share that file or commit it into Git or other
# source control.
# pylint: disable=no-name-in-module,wrong-import-order
try:
    from secrets import secrets
except ImportError:
    print("WiFi secrets are kept in secrets.py, please add them there!")
    raise

# If you are using a board with pre-defined ESP32 Pins:
esp32_cs = DigitalInOut(board.ESP_CS)
esp32_ready = DigitalInOut(board.ESP_BUSY)
esp32_reset = DigitalInOut(board.ESP_RESET)

# If you have an externally connected ESP32:
# esp32_cs = DigitalInOut(board.D9)
# esp32_ready = DigitalInOut(board.D10)
# esp32_reset = DigitalInOut(board.D5)

# If you have an AirLift Featherwing or ItsyBitsy Airlift:
# esp32_cs = DigitalInOut(board.D13)
# esp32_ready = DigitalInOut(board.D11)
# esp32_reset = DigitalInOut(board.D12)

spi = busio.SPI(board.SCK, board.MOSI, board.MISO)
[...]

PR #155, opened 2022-01-22
Allow optional nbytes parameter for recv_into

tekktrik:
That sweet, sweet hotfix for Issue #154!  Sorry for breaking some things, and thanks to @FoamyGuy to summarizing.  Default argument is None which still triggers filling buffer as much as possible, length should override that now.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -184,7 +184,7 @@ class socket:
             if avail:
                 stamp = time.monotonic()
                 recv = _the_interface.socket_read(self._socknum, min(to_read, avail))
-                # received.append(recv)
+                received.append(recv)
                 start = len(buffer) - to_read
                 to_read -= len(recv)
                 end = len(buffer) - to_read
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -161,14 +161,14 @@ class socket:
         gc.collect()
         return ret
 
-    def recv_into(self, buffer):
+    def recv_into(self, buffer, nbytes=None):
         """Read some bytes from the connected remote address into a given buffer
 
         :param bytearray buffer: The buffer to read into
         """
 
         stamp = time.monotonic()
-        to_read = len(buffer)
+        to_read = len(buffer) if nbytes is None else nbytes
         received = []
         while to_read > 0:
             # print("Bytes to read:", to_read)
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -172,7 +172,7 @@ class socket:
 
         stamp = time.monotonic()
         to_read = len(buffer)
-        limit = to_read if nbytes is None else to_read - nbytes
+        limit = 0 if nbytes is None else to_read - nbytes
         received = []
         while to_read > limit:
             # print("Bytes to read:", to_read)
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -165,12 +165,16 @@ class socket:
         """Read some bytes from the connected remote address into a given buffer
 
         :param bytearray buffer: The buffer to read into
+        :param int nbytes: (Optional) Number of bytes to receive
+            default is as many as possible before filling the
+            buffer or timing out
         """
 
         stamp = time.monotonic()
-        to_read = len(buffer) if nbytes is None else nbytes
+        to_read = len(buffer)
+        nbytes = to_read if nbytes is None else to_read - nbytes
         received = []
-        while to_read > 0:
+        while to_read > len(buffer) - nbytes:
             # print("Bytes to read:", to_read)
             avail = self.available()
             if avail:
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -177,8 +177,9 @@ class socket:
 
         stamp = time.monotonic()
         to_read = len(buffer)
+        limit = 0 if nbytes == 0 else to_read - nbytes
         received = []
-        while to_read > nbytes:
+        while to_read > limit:
             # print("Bytes to read:", to_read)
             avail = self.available()
             if avail:
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -161,20 +161,24 @@ class socket:
         gc.collect()
         return ret
 
-    def recv_into(self, buffer, nbytes=None):
+    def recv_into(self, buffer, nbytes=0):
         """Read some bytes from the connected remote address into a given buffer
 
         :param bytearray buffer: The buffer to read into
-        :param int nbytes: (Optional) Number of bytes to receive
-            default is as many as possible before filling the
+        :param int nbytes: (Optional) Number of bytes to receive default is 0,
+            which will receive as many bytes as possible before filling the
             buffer or timing out
         """
 
+        if not (0 <= nbytes <= len(buffer)):
+            raise ValueError(
+                "Can only read number of bytes between 0 and length of supplied buffer"
+            )
+
         stamp = time.monotonic()
         to_read = len(buffer)
-        limit = 0 if nbytes is None else to_read - nbytes
         received = []
-        while to_read > limit:
+        while to_read > nbytes:
             # print("Bytes to read:", to_read)
             avail = self.available()
             if avail:
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -170,7 +170,7 @@ class socket:
             buffer or timing out
         """
 
-        if not (0 <= nbytes <= len(buffer)):
+        if not 0 <= nbytes <= len(buffer):
             raise ValueError(
                 "Can only read number of bytes between 0 and length of supplied buffer"
             )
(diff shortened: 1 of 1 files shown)


adafruit_esp32spi/adafruit_esp32spi_socket.py
```diff
@@ -161,16 +161,20 @@ def recv(self, bufsize=0):
...
+            default is as many as possible before filling the
+            buffer or timing out
         """
 
         stamp = time.monotonic()
         to_read = len(buffer)
+        limit = 0 if nbytes is None else to_read - nbytes
         received = []
-        while to_read > 0:
+        while to_read > limit:
```
FoamyGuy, line 177: I think this while loop is evaluating to false and we aren't going inside of it. I added a debugging print above this to print the values of limit and to_read: `print("to_read: {} - limit: {}".format(to_read, limit))`

And I'm seeing these:
```
to_read: 32 - limit: 32
```

I also tried modifying the requests library to not pass the `nbytes` argument and it seemed to get stuck in an infinite loop, but did eventually finish after a few minutes. I think it took much longer than I was expecting based on previous experience running the simpletest. But I'm not sure if this part is actually related to this issue since requests does pass `nbytes` and that is how this is intended to work as far as I understand.

FoamyGuy (commented):
I think we still have some kind of issue related to this. When I run the requests simpletest on a PyPortal with `7.2.0-alpha-1` I am consistently getting this error raised:
```
code.py output:
Connecting to AP...
Connected to AloeVera   RSSI: -42
Fetching text from http://wifitest.adafruit.com/testwifi/index.html
  File "code.py", line 58, in <module>
  File "/lib/adafruit_requests.py", line 847, in get
  File "/lib/adafruit_requests.py", line 719, in request
  File "/lib/adafruit_requests.py", line 216, in __init__
RuntimeError: Unable to read HTTP response.

Code done running.
```
I tried several times and it always returns the same error, has never completed the request successfully for me.

I looked into this a bit during my stream today but was unable to determine the real root cause (the details of the buffer handling and internals of requests and esp32spi socket a bit over my head ATM).

FoamyGuy:
I know it wasn't modified with this PR, but I'm also curious if you know anything about why this line is commented out?

https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/52208e4752bee510b24ac775763776231b5912cc/adafruit_esp32spi/adafruit_esp32spi_socket.py#L179

Could that have been commented out accidentally? or if it's not used is there any downside to removing it entirely?

FoamyGuy:
I tried changing the while loop logic to >=:
```
while to_read >= limit:
```
When I ran it like that it seemed to get stuck in an infinite loop. I waited for several minutes, but the get request never completed (it didn't error out either just sat waiting).

anecdata:
Instead of the `limit`, would something like this work?
```py
if nbytes is a positive integer (or somesuch check):
    to_read = nbytes
else:
    to_read = len(buffer)
```

anecdata:
Or... this may be a crazy suggestion (I'm not that expert at Python), but could it work to leverage `recv` inside `recv_into`, something like:
```py
def recv_into(self, buffer, nbytes=None):
    self._buffer = buffer
    buffer = recv(nbytes)  # or 0 if nbytes is None
return len(some magic way to know how many bytes were actually put into the buffer; maybe a global)
```

tekktrik:
We could do that, but doesn't that defeat the purpose of using recv_into?  I don't know how garbage collection works, and whether the concern for memory is more in this code or the user implementation.  If that's the latter, than that's definitely the easiest way to do it, and if not at at the very least a good hotfix until it can be fleshed out.

tekktrik:
> Instead of the `limit`, would something like this work?
> 
> ```python
> if nbytes is a positive integer (or somesuch check):
>     to_read = nbytes
> else:
>     to_read = len(buffer)
> ```
> 
> And then the rest could mostly mirror `recv()`

This could work, I think we might need to change the return value to ensure it still returns the number bytes written

anecdata:
My thought (perhaps erroneously) was to use the supplied buffer, but point to it in the global `._buffer` so that the lower-level `recv` is putting bytes there and not-re-allocating. Maybe it doesn't work that way.

jasonwashburn:
> I think we still have some kind of issue related to this. When I run the requests simpletest on a PyPortal with `7.2.0-alpha-1` I am consistently getting this error raised:
> 
> ```
> code.py output:
> Connecting to AP...
> Connected to AloeVera   RSSI: -42
> Fetching text from http://wifitest.adafruit.com/testwifi/index.html
> Traceback (most recent call last):
>   File "code.py", line 58, in <module>
>   File "/lib/adafruit_requests.py", line 847, in get
>   File "/lib/adafruit_requests.py", line 719, in request
>   File "/lib/adafruit_requests.py", line 216, in __init__
> RuntimeError: Unable to read HTTP response.
> 
> Code done running.
> ```
> 
> I tried several times and it always returns the same error, has never completed the request successfully for me.
> 
> I looked into this a bit during my stream today but was unable to determine the real root cause (the details of the buffer handling and internals of requests and esp32spi socket a bit over my head ATM).

I can confirm I'm also getting the same error using this PR and several versions of example code that tries to use requests.get on a Pyportal running 7.1.1

anecdata (commented):
probably want a check that if nbytes is supplied, it's positive (and non-zero??) and isn't greater than len(buffer), just to be safe

tekktrik:
I think I found the problem:  I designed `recv_into()` to operate that if NO parameter for `nbytes` is given, that it will fill the buffer as best as possible, and if `0` is given, then it works like CPython and does nothing.  `adafruit_requests` simple test operates (with `_recv_into()` by trying to use default size of `0`.  That seems to be what's causing the issue.

However, it looks like other libraries (at least wiznet5k) operate that default of `0` for requested size translates to fill the buffer like this implementation's `None`.  I think the best way to keep consistency is probably to patch this behavior, but I think its worth considering changing the behavior to CPython's for consistency.

What does everyone else think?

Also, good catch about making sure `nbytes` is in the valid range of numbers (0 to `len(buffer)`, I can add that.

tekktrik:
Deleted my last comment because oops, I'm wrong! - https://docs.python.org/3/library/socket.html#socket.socket.recv_into

I'll change the default input of `recv_into()` to be 0 and match this (and other libraries') implementations.  I'll also add a check the number is within a valid range.

tekktrik:
Also, good catch @FoamyGuy, that line should not have been commented out!

Simple tests still won't complete but it seems to be a different issue, I think with the `adafruit_requests` library.

tekktrik:
I'm still working on why `recv_into` is causing issues for the `adafruit_requests` library but my current theory is that the issue is there.  Doing some debug to determine the issue still!

jasonwashburn:
So, I was able to get [requests_simpletest.py](https://github.com/adafruit/Adafruit_CircuitPython_Requests/blob/main/examples/requests_simpletest.py) from the requests repository examples to run without producing errors using your latest commit. However, the first request consistently takes exactly 1min, 1sec  to "complete", and only returns the first 23 chars.  I added in some prints to show the time elapsed for each request.
```
Connecting to AP...
Connected to WBurn-2G   RSSI: -40
Fetching text from http://wifitest.adafruit.com/testwifi/index.html
----------------------------------------
Text Response:  This is a test of Adafr
----------------------------------------
Time elapsed: 0:01:01
Fetching JSON data from https://httpbin.org/get
----------------------------------------
JSON Response:  {'url': 'https://httpbin.org/get', 'headers': {'User-Agent': 'Adafruit CircuitPython', 'Host': 'httpbin.org', 'X-Amzn-Trace-Id': 'Root=1-61ecb914-1f47aa991e2377d45241cddf'}, 'args': {}, 'origin': '68.13.113.212'}
----------------------------------------
Time elapsed: 0:00:05
POSTing data to https://httpbin.org/post: 31F
----------------------------------------
Data received from server: 31F
----------------------------------------
Time elapsed: 0:00:01
POSTing data to https://httpbin.org/post: {'Date': 'July 25, 2019'}
----------------------------------------
JSON Data received from server: {'Date': 'July 25, 2019'}
----------------------------------------
Time elapsed: 0:00:01

Code done running.
```

tekktrik:
Thanks for the confirmation!  I am also seeing a reduced payload response from the Adafruit-based request, but during my debugging, it does seem that it is actually getting that data but for some reason not returning it.  I'm currently digging to see why that's the case.

tekktrik:
Okay, this should be the fix!  The issue was still with how the limit of data was calculated, causing the requests simpletest to fail.  The fixes I pushed are what got me across the finish line!

anecdata:
I tested this out with some of my usual code and got confused because `recv` was still being used and `hasattr(sock, "recv_into")` was `False`. Turns out that when the Response is initialized, only some cases trigger the use of `recv_into`. For example, in the sample code in #154 only the first URL uses `recv_into`, the last three use `recv`.

In any case, the current PR code seems to work properly when `recv_into` is invoked.

FWIW, earlier I spent some time trying to get the concept of using `recv` within `recv_into` working, but got too deep in memoryview and type wrangling issues. But it later occurred to me that the Adafruit WIZnet5k library has a Python socket implementation and indeed it uses this approach. So I plugged [that](https://github.com/adafruit/Adafruit_CircuitPython_Wiznet5k/blob/fe7cff72570ecf36f853e6644d43009d1ee79c67/adafruit_wiznet5k/adafruit_wiznet5k_socket.py#L342) into ESP32SPI and it worked fine:
```py
    def recv_into(self, buf, nbytes=0, flags=0):
        """Reads some bytes from the connected remote address info the provided buffer.
        :param bytearray buf: Data buffer
        :param nbytes: Maximum number of bytes to receive
        :param int flags: ignored, present for compatibility.
        :returns: the number of bytes received
        """
        if nbytes == 0:
            nbytes = len(buf)
        ret = self.recv(nbytes)
        nbytes = len(ret)
        buf[:nbytes] = ret
        return nbytes
```

FoamyGuy (approved):
LGTM Thanks @tekktrik 

I tested the latest version successfully on PyPortal 7.2.0-alpha.1

PR #157, opened 2022-02-18
Add adafruit_requests to requirements

Neradoc:
`adafruit_esp32spi_wifimanager.py` uses `adafruit_requests`.


dhalbert (changes requested):
Thanks! Also add to https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/main/setup.py#L36

PR #160, opened 2022-03-17
Optional HTTP response on error

tekktrik:
This provides the ability to return a given HTTP response whenever an `Exception` is hit executing user code.  This allows the WSGI server to notify if a failure occurred and continue running without completely crashing.  It catches `Exception` in order to catch almost all errors since the intention is to try to keep the server alive.  If nothing is given, previous behavior is retain, which may be desirable?

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
@@ -67,12 +67,14 @@ class WSGIServer:
     A simple server that implements the WSGI interface
     """
 
-    def __init__(self, port=80, debug=False, application=None):
+    def __init__(self, port=80, debug=False, application=None, *, error_return=None):
         self.application = application
         self.port = port
         self._server_sock = socket.socket(socknum=NO_SOCK_AVAIL)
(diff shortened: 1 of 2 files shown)


dhalbert (commented):
IThanks! I am looking at `adafruit_requests` and this library and native `wifi` support on ESP32-S2, and trying to regularize the use of exceptions between the three. SO will be reviewing this, but not immediately.

tekktrik:
Sounds good!  I love the WSGI libraries so I've just playing around there lately.  Happy to changes as needed.

tekktrik:
Converting to a draft so it doesn't accidentally get merged in the meantime

tekktrik:
@dhalbert just checking in on this since it's been a few weeks, happy to keep this hibernating for now

tekktrik:
@dhalbert is this good to go or are you still working on the other libraries?  I can change this to a draft in the meantime if so.

tekktrik:
Just checking in on this @dhalbert to see if it's still relevant

tekktrik:
Based on previous discussions on how libraries like this might change, I'm closing for now.  I'll open it up again later if it makes sense.

PR #163, opened 2022-05-10
Remove this closing notice

dannystaple:
This is a little noisy. It looks like left in debug.


jepler (approved):
Thank you, looks good.

PR #164, opened 2022-05-10
Make these connection errors distinct from other error types.

dannystaple:


adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -144,7 +144,7 @@ def connect_normal(self):
                 self.esp.connect_AP(bytes(ssid, "utf-8"), bytes(password, "utf-8"))
                 failure_count = 0
                 self.pixel_status((0, 100, 0))
-            except (ValueError, RuntimeError) as error:
+            except (ValueError, OSError) as error:
```
dhalbert, line 147: A `ValueError` here now indicates something that is never going to work, so I think it should no longer be caught here, and instead should just propagate up.
```suggestion
            except OSError as error:
```

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -173,7 +173,7 @@ def create_ap(self):
                     self.esp.create_AP(bytes(self.ssid, "utf-8"), None)
                 failure_count = 0
                 self.pixel_status((0, 100, 0))
-            except (ValueError, RuntimeError) as error:
+            except (ValueError, OSError) as error:
```
dhalbert, line 176: A `ValueError` now indicates something that is never going to work, so I think it should no longer be caught here, and instead should just propagate up.
```suggestion
            except OSError as error:
```

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
```diff
@@ -203,7 +203,7 @@ def connect_enterprise(self):
                 failure_count = 0
                 self.pixel_status((0, 100, 0))
                 sleep(1)
-            except (ValueError, RuntimeError) as error:
+            except (ValueError, OSError) as error:
```
dhalbert, line 206: A `ValueError` now indicates something that is never going to work, so I think it should no longer be caught here, and instead should just propagate up.
```suggestion
            except OSError as error:
```

dhalbert (changes requested):
Thanks for improving this! The example scripts in `examples` should be updated to handle the new exceptions properly. For instance, `ValueError` probably should not be retried any more. There is probably code in Learn Guides and other libraries that will also need to be fixed.

@makermelissa FYI

This should get a major release version bump, since it is not upward compatible.

adafruit_esp32spi/adafruit_esp32spi_wifimanager.py
dannystaple: Done

dannystaple:
> Thanks for improving this! The example scripts in `examples` should be updated to handle the new exceptions properly. For instance, `ValueError` probably should not be retried any more. There is probably code in Learn Guides and other libraries that will also need to be fixed.
> 
> @makermelissa FYI
> 
> This should get a major release version bump, since it is not upward compatible.

I've updated the examples and found a couple of other Runtime errors that should be something else.

tekktrik:
The remaining CI failure was from the sphinx upgrade, re-ran it and it passed!

dhalbert (approved):
Thanks for the big rationalization of the exceptions!

PR #168, opened 2022-05-30
Examples and tests for debugging

dannystaple:


dhalbert (changes requested):
Thanks for this!

I am making some extensive changes in #167 (not yet merged): removing non-standard stuff from `socket`, and removing the `wsgi_server` stuff entirely.

dhalbert:
> Hmm - I am preparing to do some quite interesting things with the WSGI server (and possibly see about websockets, but that may be stretching too far).

I want server stuff to be outside this library, so we can use it with multiple socket implementations (ESP32-S2, etc.). So I'd really like to see https://github.com/adafruit/Adafruit_CircuitPython_WSGI be better developed. Could you see going that direction?

> I've had some stability struggles with this, that I traced into NINA-FW. [adafruit/nina-fw#46](https://github.com/adafruit/nina-fw/issues/46) - so looking at alternatives.

I did not see this report, thanks -- it may have relevance to some other problems we are seeing with ESP32SPI.

dannystaple:
> 

That sounds good indeed - it also invites other socket protocol types to be made in the same way. I am about to disappear to EMFCamp, but it might be good to discuss and perhaps see if I can pitch in teasing these apart when i am back.

FoamyGuy:
Checking in on this one,  @dannystaple are you still interested in circling back to this effort? 

It may end up being easiest to close this PR and open new one(s) if so. 
There are  a number of changes that could be complex with the current state of this PR:

- Extensive changes within esp32spi that are merged now that Dan mentioned in a prior comment. Among other things this changed the exception types that were used which appear to be several of the changes in this PR as well.
- The WSGI server components now live in: https://github.com/adafruit/Adafruit_CircuitPython_WSGI/tree/main/adafruit_wsgi so examples related to that can be added over in that repo. 
- For non-esp32spi based network there is also https://github.com/adafruit/Adafruit_CircuitPython_HTTPServer which can be used for implementing a basic web server, **_but this library cannot work with esp32spi_** only core `wifi` module or CPython sockets are supported.

dannystaple:
Hello, I think that makes sense at the moment.Sent from my iPhoneOn 7 Aug 2023, at 21:50, foamyguy ***@***.***> wrote:﻿
Checking in on this one,  @dannystaple are you still interested in circling back to this effort?
It may end up being easiest to close this PR and open new one(s) if so.
There are  a number of changes that could be complex with the current state of this PR:

Extensive changes within esp32spi that are merged now that Dan mentioned in a prior comment. Among other things this changed the exception types that were used which appear to be several of the changes in this PR as well.
The WSGI server components now live in: https://github.com/adafruit/Adafruit_CircuitPython_WSGI/tree/main/adafruit_wsgi so examples related to that can be added over in that repo.
For non-esp32spi based network there is also https://github.com/adafruit/Adafruit_CircuitPython_HTTPServer which can be used for implementing a basic web server, but this library cannot work with esp32spi only core wifi module or CPython sockets are supported.

PR #169, opened 2022-06-07
fix recv_into again

dhalbert:
My original fix in #166 did not handle the case of more bytes arriving while in the `while` loop in `socket.recv_into()`. I discovered this while trying to debug something else.

@FoamyGuy Would you be willing to test this? It is unrelated to your fix for unlimited `recv()`.


FoamyGuy (approved):
I tested the simpletest from this repo and the Cleveland Art Museum project on a PyPortal Titano using the branch from this PR and both are working well.

PR #173, opened 2022-08-31
Add optional requirements to file

### LINKED ISSUE #172: Missing optional requirements
This library has examples using external libraries that are not contained within `requirements.txt`.  These libraries should be added to `optional_requirements.txt` so they can be installed if needed when working with these examples.

tekktrik:
Resolves #172


FoamyGuy (approved):
Looks good to me. Thanks for the fix @tekktrik

PR #175, opened 2022-09-10
Close #174 by sending entire webserver response to ESP32 in one go if possible

### LINKED ISSUE #174: 800 bps when serving webpage with WSGIServer
Hi! I've been messing around with a Metro M4 Airlift Lite and hit an interesting snag when trying to serve a webpage using Adafruit's `esp32spi_wsgiserver` and `adafruit_wsgi.wsgi_app`. (Code is [here](https://github.com/rivques/CircuitPython/blob/master/internet.py) for those interested.) When serving the webpage, I was getting only about 800 *bits* per second, slow enough to cause Chrome to have issues recognizing the format of the page as HTML because (I assume) the `<!DOCTYPE html>` did not arrive quickly enough. After digging through `debug=True` logs and the various modules of `adafruit_esp32spi`, it seems like the issue is that `WSGIServer.finish_response` is sending data to `ESP_SPIcontrol.socket_write` one byte at a time. This bypasses `socket_write`'s built-in chunking and seems to cause a lot of extra overhead, possibly due to #108. To fix this I use a [slightly modified version](https://github.com/rivques/CircuitPython/blob/master/lib/my_wsgi.py) of `esp32spi_wsgiserver` where I `socket_write` the entire response at once if possible instead of looping over every char in the response and writing it individually. This increased data rate to something like 178 kbps. Anyways, my questions are:
1. Am I missing something that would cause this to be a non-issue?
1. Is this a known issue and if so, is there another workaround?
2. Would this problem be fixed by #108 being resolved?
3. Is my workaround a reasonable one, or does it have pitfalls I can't see? If so, should I put it in a PR to this repo?

rivques:
This fixes #174, and gets my connection speed on a Metro M4 Airlift Lite from about 800 bps to something like 178 kpbs when serving a webpage, though your mileage may vary. For some reason `finish_response` would be called an empty array after the initial call with the intended response string, so I've left the original handling logic behind some type checks if that array ends up being non-empty. It's very possible I've misunderstood how WSGI is working here and more needs to be done with this, but it seems to handle what I've thrown at it.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_wsgiserver.py
@@ -119,11 +119,11 @@ class WSGIServer:
                 response += "{0}: {1}\r\n".format(*header)
             response += "\r\n"
             self._client_sock.send(response.encode("utf-8"))
-            if isinstance(result, bytes): # send whole response if possible (see #174)
+            if isinstance(result, bytes):  # send whole response if possible (see #174)
                 self._client_sock.send(result)
             elif isinstance(result, str):
                 self._client_sock.send(result.encode("utf-8"))
-            else: # fall back to sending byte-by-byte
+            else:  # fall back to sending byte-by-byte
                 for data in result:
                     if isinstance(data, bytes):
                         self._client_sock.send(data)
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
This is fine with me for now, for the reasons noted in #174. Thanks!

dhalbert:
The checks failed, only due to a formatting issue. Could you run your changes through `black`? Or use pre-commit to do it automatically.

PR #178, opened 2023-04-26
Don't print password by default in debug

### LINKED ISSUE #156: Prints password in clear text if debug is True
This library prints the wifi password from `secrets.py` in cleartext if debug is True from the code here:

https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/blob/52208e4752bee510b24ac775763776231b5912cc/adafruit_esp32spi/adafruit_esp32spi.py#L531

I can see how this could be helpful for troubleshooting, especially noticing if the password was incorrect. But I think it should require an extra boolean parameter or something to do this. 

I would like to see all the rest of the debugging messages printed, but don't really want my password printed along with it.

zachariahpifer:
closes #156

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -325,7 +325,7 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
         *,
         reply_params=1,
         sent_param_len_16=False,
-        recv_param_len_16=False
+        recv_param_len_16=False,
     ):
         """Send a high level SPI command, wait and return the response"""
         self._send_command(cmd, params, param_len_16=sent_param_len_16)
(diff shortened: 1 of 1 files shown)


adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -563,17 +563,23 @@ def connect(self, secrets):
         that contains a 'ssid' and 'password' entry"""
         self.connect_AP(secrets["ssid"], secrets["password"])
 
-    def connect_AP(self, ssid, password, timeout_s=10):  # pylint: disable=invalid-name
+    def connect_AP(
+        self, ssid, password, timeout_s=10, show_password=False
```
tannewt, line 567: I'd rather not have the kwarg here. I think it should go next to the `debug` flag. It could be `debug_show_secrets` to cover other sorts of keys in addition to the password.

tannewt (changes requested):
Thank you for the PR! One suggestion.

adafruit_esp32spi/adafruit_esp32spi.py
zachariahpifer: Good thinking! To be clear, you're suggesting it should be an __init__ param?

FoamyGuy (approved):
Thanks for working on this @zachariahpifer!

The latest version of this looks good to me. I haven't tested on hardware. Will try it out in the next few days if it's still open.

PR #182, opened 2023-11-03
Another settings.toml update, slightly different approach

RetiredWizard:
I decided to take a stab at reworking the examples in this library to work with settings.toml and didn't notice the existing pr #179.

I took a slightly different approach as I didn't actually make any functional changes to the core library. The library currently doesn't directly access the settings.py file but uses a dictionary that is passed in and I decided it was less breaking to simply modify the examples to build the passed in dictionary from settings.toml. I also left a fallback to reading the secrets.py file in the examples because I didn't know when/if any learn guides that might reference the examples would be updated.

I considered just tossing this out, once I spotted the existing PR but since the work was already done, I figured I'd throw it out there anyway. There are advantages to modifying the library to access the secrets directly from the settings.toml environment variable, particularly with regards to simplifying the usage of library and example programs, so I'm not sure my approach makes sense. 

edit: I tested most of the example programs using an Arduino Nano Connect RP2040.


RetiredWizard:
It has occurred to me that having a basic utility to create a secrets dictionary from the settings.toml environment parameters would ease the process of eliminating the secrets.py file. If such a function existed then moving to settings.toml would simply require the replacement of `import secrets.py` with an assignement similar to `secrets = os.build_secrets()`.  Just a though....

FoamyGuy (approved):
I do like this method of supporting settings.toml better since it eases the transition to the new way. Re-using the dictionary secrets that has been passed in is nice too because that can be utilized in the CPython + Blinka Context which can't use secrets.py directly due to naming collision with the builtin.

I do also like the idea of some conversion tool that can convert toml data into a dictionary as a means to bridge the distance between the two ways of configuration

PR #183, opened 2024-01-15
Fix double subtraction of bytes_read from num_to_read in recv_into

us3r64:


dhalbert:
Thanks for this fix! So this fixes the problem you were seeing, in your experience?

dhalbert (approved):
Thank you!

PR #185, opened 2024-01-15
Fix potential socket data read already into self._buffer

### LINKED ISSUE #186: Fix WSGI headers parse and readline implementation
Socket data might be already read into `self._buffer`, especially with https://github.com/adafruit/Adafruit_CircuitPython_WSGI/blob/main/adafruit_wsgi/esp32spi_wsgiserver.py#L67 implementation.

Reading everything from socket is going to include HTTP body also but discarded if `self._buffer` is not rechecked.

Fix: https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/pull/185

us3r64:
Fixes #186

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_socket.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socket.py
@@ -113,6 +113,17 @@ class socket:
         num_to_read = len(buffer) if nbytes == 0 else nbytes
         num_read = 0
         while num_to_read > 0:
+            # we might have read socket data into the self._buffer with:
+            # esp32spi_wsgiserver: socket_readline
+            if len(self._buffer) > 0:
+                bytes_to_read = min(num_to_read, len(self._buffer))
+                buffer[num_read : num_read + bytes_to_read] = self._buffer[:bytes_to_read]
+                num_read += bytes_to_read
+                num_to_read -= bytes_to_read
+                self._buffer = self._buffer[bytes_to_read:]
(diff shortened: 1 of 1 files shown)


adafruit_esp32spi/adafruit_esp32spi_socket.py
```diff
@@ -113,6 +113,19 @@ def recv_into(self, buffer, nbytes: int = 0):
         num_to_read = len(buffer) if nbytes == 0 else nbytes
         num_read = 0
         while num_to_read > 0:
+            # we might have read socket data into the self._buffer with:
+            # esp32spi_wsgiserver: socket_readline
```
dhalbert, line 117: We discarded `esp32spi_wsigiserver`. Are there other circumstances where is could happen?

us3r64: I haven't checked any other libraries where `_buffer` is used, only https://github.com/adafruit/Adafruit_CircuitPython_WSGI/blob/main/adafruit_wsgi/esp32spi_wsgiserver.py

But i think is kinda of expected behavior and reasonable to use the `_buffer`, read as much as possible from the socket, because is not performant to read one bit at a time to find EOL for example.

This implementation also made into Adafruit_CircuitPython_Wiznet5k,  adafruit_wiznet5k_socket `_readline` and adafruit_wiznet5k_wsgiserver so maybe is spread.

dhalbert (approved):

PR #192, opened 2024-02-29
tweak output formatting

mMerlin:
A few tweaks to pretty up the output formatting.

- Convert the version byte array to a utf-8 string
- display the mac address as colon separated hex byte values, instead of a list
- use a fixed width field for ssid values, so that the RSSI lines up better
  - if want to handle ssid up to the maximum 32 bytes, I suggest replacing the leading tab character with a couple of spaces.


dhalbert (approved):
Thanks!

PR #198, opened 2024-04-25
Convert to SocketPool

justmobilize:
For use with: https://github.com/adafruit/Adafruit_CircuitPython_ConnectionManager/pull/11

Converting from the old `adafruit_esp32spi/adafruit_esp32spi_socket.py` to `adafruit_esp32spi/adafruit_esp32spi_socketpool.py` to remove `set_interface` and work more like a built-in SocketPool.

This also allows one to use multiple ESP32SPI boards at the same time.

General example:
```py
import board
import busio
import digitalio
import os
import adafruit_requests
from adafruit_esp32spi.adafruit_esp32spi import ESP_SPIcontrol
import adafruit_esp32spi.adafruit_esp32spi_socketpool as socketpool

TEXT_URL = "http://wifitest.adafruit.com/testwifi/index.html"
wifi_ssid = os.getenv("CIRCUITPY_WIFI_SSID")
wifi_password = os.getenv("CIRCUITPY_WIFI_PASSWORD")

esp32_cs = digitalio.DigitalInOut(board.D13)
esp32_ready = digitalio.DigitalInOut(board.D11)
esp32_reset = digitalio.DigitalInOut(board.D12)
spi_bus = busio.SPI(board.SCK, MOSI=board.MOSI, MISO=board.MISO)
radio = ESP_SPIcontrol(spi_bus, esp32_cs, esp32_ready, esp32_reset)
radio.connect_AP(wifi_ssid, wifi_password)
pool = socketpool.SocketPool(radio)
requests = adafruit_requests.Session(pool)
print(requests.get(TEXT_URL).text)
```

or with the new `ConnectionManager`
```py
import board
import busio
import digitalio
import os
import adafruit_connection_manager
import adafruit_requests
from adafruit_esp32spi.adafruit_esp32spi import ESP_SPIcontrol
from adafruit_esp32spi import adafruit_esp32spi_socketpool

TEXT_URL = "http://wifitest.adafruit.com/testwifi/index.html"
wifi_ssid = os.getenv("CIRCUITPY_WIFI_SSID")
wifi_password = os.getenv("CIRCUITPY_WIFI_PASSWORD")

esp32_cs = digitalio.DigitalInOut(board.D13)
esp32_ready = digitalio.DigitalInOut(board.D11)
esp32_reset = digitalio.DigitalInOut(board.D12)
spi_bus = busio.SPI(board.SCK, MOSI=board.MOSI, MISO=board.MISO)
[...]

PR #201, opened 2024-05-06
socket.connect(): Auto-select UDP_MODE

jepler:
This improves compatibility with the standard socket module

Not tested. Prompted by https://github.com/adafruit/Adafruit_CircuitPython_NTP/pull/32

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
@@ -121,7 +121,7 @@ class Socket:
 
     def send(self, data):
         """Send some data to the socket."""
-        if self._type is SocketPool.SOCK_DGRAM:
+        if self._type == SocketPool.SOCK_DGRAM:
             conntype = self._interface.UDP_MODE
         else:
             conntype = self._interface.TCP_MODE
--- a/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
@@ -108,7 +108,11 @@ class Socket:
         depending on the underlying interface"""
         host, port = address
         if conntype is None:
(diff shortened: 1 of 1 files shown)


justmobilize:
@jepler this makes sense to me

dhalbert (approved):
Thanks! Tested by https://github.com/adafruit/Adafruit_CircuitPython_NTP/pull/32#issuecomment-2096018649.

FoamyGuy:
Does this need to come along with some other change inside adafruit_connection_manager or elsewhere?  I tried this branch on a PyPortal Titano with the currently released connection manager and the esp32spi_simpletest.py from requests library but I seem to consistently get this exception raised:

```
  File "code.py", line 46, in <module>
  File "adafruit_connection_manager.py", line 124, in get_radio_socketpool
ImportError: no module named 'adafruit_esp32spi.adafruit_esp32spi_socket'

```
Maybe this is related to https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/pull/198

FoamyGuy:
> Does this need to come along with some other change...

I understand now. `adafruit_connection_manager` is frozen into the builds it appears. And the current release of ESP32SPI is not compatible with the version of `adafruit_connection_manager` that is frozen in to the current stable release (haven't tried newest). 

With the current connection manager pasted into the root of CIRCUITPY to override the frozen instance, I was able to test this successfully as well with that simpletest script.

As an aside, it might be nice if circup were aware of the list of frozen modules per device so that it could give you a warning if / when you install a library that is not going to have any effect by default due to being overridden by a frozen one.

justmobilize:
@FoamyGuy I had asked at one point about frozen libs, and since the amount of devices were low, it was said it wasn't worth it.

Also, it will install things in lib that are frozen, and then they don't get used... I had thought of adding a method to install in root to fix this as well.

Happy to take this in if wanted

PR #202, opened 2024-05-15
Add sendto

justmobilize:
Helps fix: https://github.com/adafruit/Adafruit_CircuitPython_NTP/issues/33


dhalbert (approved):
OK, let's do it, and we can change something if it doesn't work for other than NTP.

b-blake:
Version 3.1.1 works fine on my Raspberry Pi Pico W.
Thanks for the fix.

PR #204, opened 2024-05-20
Fix SocketPoolConstants typo

justmobilize:
I can't spell. Another user found a type in [WIZnet](https://github.com/adafruit/Adafruit_CircuitPython_Wiznet5k/pull/161), so also fixing it here

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
@@ -29,7 +29,7 @@ from adafruit_esp32spi import adafruit_esp32spi as esp32spi
 _global_socketpool = {}
 
 
-class SocketPoolContants:  # pylint: disable=too-few-public-methods
+class SocketPoolConstants:  # pylint: disable=too-few-public-methods
     """Helper class for the constants that are needed everywhere"""
 
     SOCK_STREAM = const(0)
@@ -40,7 +40,7 @@ class SocketPoolContants:  # pylint: disable=too-few-public-methods
     MAX_PACKET = const(4000)
 
 
-class SocketPool(SocketPoolContants):
+class SocketPool(SocketPoolConstants):
     """ESP32SPI SocketPool library"""
 
     def __new__(cls, iface: ESP_SPIcontrol):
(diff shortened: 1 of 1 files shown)


dhalbert (changes requested):
I am thinking that the superclass `SocketPoolConstants` is not really needed, since the native `socketpool` doesn't have it either. It would save space and time to just have the constants as class attributes in `SocketPool`.

adafruit_esp32spi/adafruit_esp32spi_socketpool.py
```diff
@@ -60,12 +60,12 @@ def getaddrinfo(  # pylint: disable=too-many-arguments,unused-argument
...
             raise ValueError("Port must be an integer")
         ipaddr = self._interface.get_host_by_name(host)
-        return [(SocketPoolContants.AF_INET, socktype, proto, "", (ipaddr, port))]
+        return [(SocketPoolConstants.AF_INET, socktype, proto, "", (ipaddr, port))]
 
     def socket(  # pylint: disable=redefined-builtin
         self,
-        family=SocketPoolContants.AF_INET,
-        type=SocketPoolContants.SOCK_STREAM,
+        family=SocketPoolConstants.AF_INET,
```
justmobilize, line 67: @dhalbert I'm open to suggestions, but this is why the superclass was added. Since we are in the class, it's not defined and thus this line will error on import:
image
_ignore the line number in the image, it's this one_

dhalbert: I see. But what about something like this? This works:
```py
class ConstantsTest:
    A = 1
    B = 2

    def f(self, a=A, b=B):
        pass
```
The unqualified name works.

This actually is more complicated. In CPython, `AF_INET`, etc. are defined at the module level, not inside the `Socket` class. I can't remember why we defined them as class constants in `SocketPool` instead of in `socketpool`.

justmobilize: Ahh, I can do a commination of both...

I'm assuming they are in `SocketPool`, since that's where they were when ESP32SPI was created and so when `socketpool.SocketPool` was created if followed suit?

PR #205, opened 2024-05-20
Remove SocketPoolContants

justmobilize:
Remove mis-spelled `SocketPoolContants` and place the constants where they should be

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
@@ -29,8 +29,8 @@ from adafruit_esp32spi import adafruit_esp32spi as esp32spi
 _global_socketpool = {}
 
 
-class SocketPoolContants:  # pylint: disable=too-few-public-methods
-    """Helper class for the constants that are needed everywhere"""
+class SocketPool:
+    """ESP32SPI SocketPool library"""
 
     SOCK_STREAM = const(0)
     SOCK_DGRAM = const(1)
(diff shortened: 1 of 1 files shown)


justmobilize:
@dhalbert closed #204 and re-opened

adafruit_esp32spi/adafruit_esp32spi_socketpool.py
```diff
@@ -60,12 +56,12 @@ def getaddrinfo(  # pylint: disable=too-many-arguments,unused-argument
         if not isinstance(port, int):
             raise ValueError("Port must be an integer")
         ipaddr = self._interface.get_host_by_name(host)
-        return [(SocketPoolContants.AF_INET, socktype, proto, "", (ipaddr, port))]
+        return [(SocketPool.AF_INET, socktype, proto, "", (ipaddr, port))]
```
dhalbert, line 59: This should not be a qualified name now.
```suggestion
        return [(AF_INET, socktype, proto, "", (ipaddr, port))]
```

dhalbert (changes requested):

adafruit_esp32spi/adafruit_esp32spi_socketpool.py
justmobilize: @dhalbert happy to add this and the pylint ignore, but thought this was cleaner (and it works):
image

dhalbert: OK, I see, sorry. I was just testing by importing it into CPython, which did not complain. But it doesn't run.

dhalbert (approved):
This is cleaner, yes.

PR #206, opened 2024-05-21
Remove timeout Exception

justmobilize:
This removes the custom timeout Exception that makes this socket library behave like CPython versions pre 3.10 and updates it to behave like espressif and pico-w.

This should fix the issues with https://github.com/adafruit/Adafruit_CircuitPython_MiniMQTT/issues/216

If this is successful, a similar PR can be opened for the WIZnet5k, and a comment only PR for MiniMQTT to specify what devices hit which paths

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
+++ b/adafruit_esp32spi/adafruit_esp32spi_socketpool.py
@@ -21,6 +21,7 @@ except ImportError:
     pass
 
 
+import errno
 import time
 import gc
 from micropython import const
@@ -181,7 +182,7 @@ class Socket:
                 break
             # No bytes yet, or more bytes requested.
             if self._timeout > 0 and time.monotonic() - last_read_time > self._timeout:
-                raise timeout("timed out")
+                raise OSError(errno.ETIMEDOUT)
         return num_read
 
     def settimeout(self, value):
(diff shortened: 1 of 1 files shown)


justmobilize:
@prcutler I did some basic testing with this and didn't have any issues. Would you be willing to test? In order to do so, can you:

1. Install MiniMQTT v7.8.0 (which is now in the bundle)
2. Download this version on esp32spi and place it in the root (not the lib, since it's frozen on the pyportal)
3. Run your code

If all goes well this should work as planned. I would let it loop through enough times that it logs `KeepAlive period elapsed`, then try to post a message and see if it keeps going through.

dhalbert:
Copying the whole adafruit_esp32spi/ directory to CIRCUITPY is correct. Then also remove the solo files you copied over.

prcutler:
This now is succesfully working - thanks @justmobilize .

I've sent a few messages without any crashes and it's been up and running for the last 15 minutes without any issues.

dhalbert (approved):
This looks good and thanks @prcutler for testing.

Do you want to clean up the code or comments in MiniMQTT? I think it should still handle the `timeout` exception, but the comments might be misleading. It could also handle `TimeOutError` explicitly. `TimeoutError` was added in CPython 3.3 so I think it's fine to use it.

PR #207, opened 2024-05-23
Clear error for only one SSL socket open at a time

justmobilize:


justmobilize:
@dhalbert tiny change, will save future hours of debugging

PR #209, opened 2024-06-12
Add ap_info

justmobilize:
This moves `ssid`, `bssid` and `rssi` to a new network class to match `wifi.radio`. Also added in `authmode` since the call was there, but not being used.

Also made `firmware_version` a string, since it always needs to be formated.

Once Approved, will open PRs examples in other libraries and guides.

### DIFF
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -195,9 +195,7 @@ class ESP_Network:
     @property
     def country(self):
         """String id of the country code"""
-        if self._raw_country:
-            return self._raw_country
-        return None
+        return self._raw_country
 
     @property
     def authmode(self):
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -131,7 +131,7 @@ ADC_ATTEN_DB_11 = const(3)
 # pylint: disable=too-many-lines
 
 
-class ESP_Network:
+class Network:
     """A wifi network provided by a nearby access point."""
 
     def __init__(  # pylint: disable=too-many-arguments
@@ -491,7 +491,7 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
             channel = self._send_command_get_response(_GET_IDX_CHAN_CMD, ((i,),))[0]
             authmode = self._send_command_get_response(_GET_IDX_ENCT_CMD, ((i,),))[0]
             APs.append(
-                ESP_Network(
+                Network(
                     raw_ssid=name,
                     raw_bssid=bssid,
                     raw_rssi=rssi,
@@ -608,7 +608,7 @@ class ESP_SPIcontrol:  # pylint: disable=too-many-public-methods, too-many-insta
         """Network object containing BSSID, SSID, authmode, channel, country and RSSI when
         connected to an access point. None otherwise."""
         if self.is_connected:
-            return ESP_Network(esp_spi_control=self)
+            return Network(esp_spi_control=self)
         return None
 
     @property
--- a/adafruit_esp32spi/adafruit_esp32spi.py
+++ b/adafruit_esp32spi/adafruit_esp32spi.py
@@ -131,6 +131,97 @@ ADC_ATTEN_DB_11 = const(3)
 # pylint: disable=too-many-lines
 
 
+class ESP_Network:
+    """A wifi network provided by a nearby access point."""
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        esp_spi_control=None,
+        raw_ssid=None,
+        raw_bssid=None,
+        raw_rssi=None,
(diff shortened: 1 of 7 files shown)


adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -131,6 +131,97 @@
 # pylint: disable=too-many-lines
 
 
+class ESP_Network:
```
justmobilize, line 134: Built this to match `wifi.radio` the best I could

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -359,7 +450,7 @@ def firmware_version(self):
         if self._debug:
             print("Firmware version")
         resp = self._send_command_get_response(_GET_FW_VERSION_CMD)
-        return resp[0]
+        return resp[0].decode("utf-8").replace("\x00", "")
```
justmobilize, line 453: everyone does this, so fixed it to a string

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -397,16 +488,19 @@ def get_scan_networks(self):
...
             bssid = self._send_command_get_response(_GET_IDX_BSSID_CMD, ((i,),))[0]
-            a_p["bssid"] = bssid
-            chan = self._send_command_get_response(_GET_IDX_CHAN_CMD, ((i,),))[0]
-            a_p["channel"] = chan[0]
-            APs.append(a_p)
+            rssi = self._send_command_get_response(_GET_IDX_RSSI_CMD, ((i,),))[0]
+            channel = self._send_command_get_response(_GET_IDX_CHAN_CMD, ((i,),))[0]
+            authmode = self._send_command_get_response(_GET_IDX_ENCT_CMD, ((i,),))[0]
+            APs.append(
+                ESP_Network(
```
justmobilize, line 496: Use new class

justmobilize:
@dhalbert would love your thoughts on this

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -131,6 +131,97 @@
...
+        if self._raw_channel:
+            return self._raw_channel[0]
+        return None
+
+    @property
+    def country(self):
+        """String id of the country code"""
+        if self._raw_country:
+            return self._raw_country
+        return None
```
dhalbert, line 200: If self._raw_country is false-y, then it's None, so the `if` is unnecessary, right?
```suggestion
        return self._raw_country
```

adafruit_esp32spi/adafruit_esp32spi.py
dhalbert: Maybe call this `Radio`?

dhalbert (changes requested):
I like the idea of making a `Radio`-like` helper class. It will take up a little extra space when frozen or imported. I hope that won't be a problem on, say, Matrix Portal M4, but I think.

adafruit_esp32spi/adafruit_esp32spi.py
```diff
@@ -131,6 +131,97 @@
...
+
+        if response == 7:
+            return "OPEN"
+        if response == 5:
+            return "WEP"
+        if response == 2:
+            return "PSK"
+        if response == 4:
+            return "WPA2"
+        return "UNKNOWN"
```
dhalbert, line 222: An aside: As a size experiment, I extracted this code into a little standalone program, and tried the same thing, but with a dict. Interestingly, the dict code was  veryslightly bigger, in terms of .mpy size. When I add three more entries, then the dict was slightly smaller.

adafruit_esp32spi/adafruit_esp32spi.py
justmobilize: I tried to match CP WiFi, and there it's called `Network`: https://docs.circuitpython.org/en/latest/shared-bindings/wifi/index.html#wifi.Network

adafruit_esp32spi/adafruit_esp32spi.py
justmobilize: Oh that's interesting!

justmobilize: I did have the thought once the Nina firmware is updated, to change that to return a string instead of the random int value

adafruit_esp32spi/adafruit_esp32spi.py
justmobilize: @dhalbert do you think just `Network`?

dhalbert: Yep, that's exactly what I was going to say. If it quacks like a `Duck`, it doesn't have to be an `ESP_Duck`.

justmobilize: Fixed here: 2da98438cf81145166f89cd7fa3969e32ca186af

dhalbert (approved):
:+1:

PR #217, opened 2025-02-21
Use adafruit-hosted sample JSON URL

dhalbert:
The sample HTTP JSON URL in a couple of tests was on http://api.coindesk.com, which doesn't exist anymore. Instead, use the (newly added) URL: http://wifitest.adafruit.com/testwifi/sample.json.

Tested on a Metro M4 AirLift Lite.

The examples are used on Learn Guide pages, so a quick review would be good, since right now the examples are broken.


dhalbert:
Merging myself, because I need to update the guide pages.

